PR #53, opened 2026-03-13
Add stackit-pod-identity-webhook image and configuration

jastBytes:

/kind enhancement

The changes introduce a new Pod Identity Webhook component for STACKIT cloud provider, adding a new chart for the webhook and its associated resources (Deployment, Service, RBAC, MutatingWebhookConfiguration). The webhook is configured to validate pod identities and enforce workload identity. The changes are integrated into the control plane and shoot system components, with proper configuration and error handling.

**Breaking changes**:

No breaking changes.


imagevector/images.yaml
```diff
@@ -135,3 +135,6 @@ images:
 - name: stackit-alb-controller-manager
   repository: reg3.infra.ske.eu01.stackit.cloud/temp/alb-controller-manager
   tag: "1245"
+- name: stackit-pod-identity-webhook
+  repository: reg3.infra.ske.eu01.stackit.cloud/stackitcloud/stackit-pod-identity-webhook
+  tag: "726f2f0@sha256:fca1f67cd7e6a515e795a34ae45d0c239379d051e494dc202033f6987b41b154"
```
timebertt: For the record (as discussed in chat): the `stackit-pod-identity-webhook` repository should be public and released before merging this integration PR in the extension.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/rbac.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: v1
```
timebertt: As mentioned in another comment: we certainly do not need a `ServiceAccount` in the seed cluster for this component. The `ServiceAccount` in the shoot cluster should be created via a shoot access secret.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
```
timebertt: Should we make this service topology-aware? ref https://github.com/gardener/gardener/blob/master/docs/development/component-checklist.md#high-availability--stability

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -1273,3 +1291,36 @@ func cleanupCloudProviderConfigSecret(ctx context.Context, client k8sclient.Clie
...
+	secretsReader secretsmanager.Reader,
+	scaledDown bool,
+) (map[string]any, error) {
+	tlsSecret, found := secretsReader.Get(stackitPodIdentityWebhookServerName)
+	if !found {
+		return nil, fmt.Errorf("secret %q not found", stackitPodIdentityWebhookServerName)
+	}
+
+	return map[string]any{
+		"replicaCount": extensionscontroller.GetControlPlaneReplicas(cluster, scaledDown, 2),
```
timebertt: Nit: typically `1` is specified here, because the high availability config webhook will manage the replica count

timebertt (changes requested):
Thanks for the PR, looking forward to this feature :)

jastBytes:
Post review changes implemented and tested via ondemand. All working now. One question open from my point of view. See comment in code.

crigertg (dismissed):

charts/internal/shoot-system-components/charts/stackit-pod-identity-webhook/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -1,33 +1,27 @@
...
         resources: ["pods"]
     admissionReviewVersions: ["v1"]
     sideEffects: None
-    failurePolicy: {{ .Values.webhook.failurePolicy | default "Ignore" }}
+    failurePolicy: Fail
     namespaceSelector:
       matchExpressions:
         - key: kubernetes.io/metadata.name
           operator: NotIn
           values: ["kube-system", "garden"]
```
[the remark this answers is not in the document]
jastBytes: Initially I was thinking the same to only target customer shoots, but @timebertt suggested that this feature should be available for ourselves and there is no real need to exclude seeds. I am not "sattelfest" enough to have a strong opinion on this topic though.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: stackit-pod-identity-webhook
+        workload-identity.stackit.cloud/skip-pod-identity-webhook: "true"
+        gardener.cloud/role: controlplane
+        networking.gardener.cloud/to-dns: allowed
+        networking.resources.gardener.cloud/to-kube-apiserver-tcp-443: allowed
+    spec:
+      topologySpreadConstraints:
```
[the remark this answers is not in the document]
jastBytes: You're right. This should be set automatically by Gardener via the label `high-availability-config.resources.gardener.cloud/type` that is already present on the deployment.

charts/internal/shoot-system-components/charts/stackit-pod-identity-webhook/templates/mutatingwebhookconfiguration.yaml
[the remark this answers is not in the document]
timebertt: We can probably leave out the `garden` exclusion. We should simply ensure that the webhook config is compliant with gardenlet's webhook checker/remediator: https://github.com/gardener/gardener/blob/master/docs/usage/shoot/shoot_status.md#constraints /

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/vpa.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: stackit-pod-identity-webhook
+  namespace: {{ .Release.Namespace }}
+spec:
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: stackit-pod-identity-webhook
+  updatePolicy:
+    updateMode: Auto
```
timebertt:
```suggestion
    updateMode: InPlaceOrRecreate
```

ref https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/docs/api.md#updatemode

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/vpa.yaml
```diff
@@ -0,0 +1,21 @@
...
+  resourcePolicy:
+    containerPolicies:
+    - containerName: stackit-pod-identity-webhook
+      minAllowed:
+        memory: 80M
+      maxAllowed:
+        cpu: {{ .Values.vpa.resourcePolicy.maxAllowed.cpu }}
+        memory: {{ .Values.vpa.resourcePolicy.maxAllowed.memory }}
+      controlledValues: RequestsOnly
```
timebertt, line 22: When scaling requests only, the configured limits will cap off even before hitting `maxAllowed`. 
Consider removing the memory limit or aligning it with the `maxAllowed` values.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/vpa.yaml
```diff
@@ -0,0 +1,21 @@
...
+    apiVersion: apps/v1
+    kind: Deployment
+    name: stackit-pod-identity-webhook
+  updatePolicy:
+    updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+    - containerName: stackit-pod-identity-webhook
+      minAllowed:
+        memory: 80M
```
timebertt: Shouldn't this be `80Mi`?

charts/internal/seed-controlplane/requirements.yaml
```diff
@@ -19,3 +19,7 @@ dependencies:
   repository: http://localhost:10191
   version: 0.1.0
   condition: stackit-alb-controller-manager.enabled
+- name: stackit-pod-identity-webhook
+  repository: http://localhost:10191
+  version: 0.1.0
+  condition: stackit-pod-identity-webhook.enabled
```
timebertt, line 25: Can you add this value to the `values.yaml` file for completeness?

charts/internal/shoot-system-components/requirements.yaml
```diff
@@ -15,3 +15,7 @@ dependencies:
   repository: http://localhost:10191
   version: 0.1.0
   condition: stackit-blockstorage-csi-driver.enabled
+- name: stackit-pod-identity-webhook
+  repository: http://localhost:10191
+  version: 0.1.0
+  condition: stackit-pod-identity-webhook.enabled
```
timebertt, line 21: Can you add this value to the `values.yaml` file for completeness?

charts/internal/shoot-system-components/charts/stackit-pod-identity-webhook/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: stackit-pod-identity-webhook
+  labels:
+    app.kubernetes.io/name: stackit-pod-identity-webhook
+webhooks:
+  - name: stackit-pod-identity-webhook.stackit.cloud
+    clientConfig:
+      url: {{ .Values.webhook.url | quote }}
```
timebertt: Can you add this value to the `values.yaml` file for completeness?

charts/internal/shoot-system-components/charts/stackit-pod-identity-webhook/values.yaml
```diff
@@ -0,0 +1,3 @@
+webhook:
+  caBundle: "" # will be set by valuesprovider
+  controlPlaneNamespace: "" # will be set by valuesprovider
```
timebertt: Seems to be unused

charts/internal/shoot-system-components/charts/stackit-pod-identity-webhook/templates/mutatingwebhookconfiguration.yaml
timebertt: Please trim trailing whitespaces

imagevector/images.yaml
```diff
@@ -135,3 +135,6 @@ images:
 - name: stackit-alb-controller-manager
   repository: reg3.infra.ske.eu01.stackit.cloud/temp/alb-controller-manager
   tag: "1245"
+- name: stackit-pod-identity-webhook
+  repository: ghcr.io/stackitcloud/stackit-pod-identity-webhook
+  tag: "v0.1.0@sha256:5c124efb00b5b5e0fd64b635e1643cebbd407b4abafdb966e02639627f41631c"
```
timebertt: Is renovate able to update this reference including the digest?

imagevector/images.yaml
```diff
@@ -135,3 +135,6 @@ images:
 - name: stackit-alb-controller-manager
   repository: reg3.infra.ske.eu01.stackit.cloud/temp/alb-controller-manager
   tag: "1245"
+- name: stackit-pod-identity-webhook
+  repository: ghcr.io/stackitcloud/stackit-pod-identity-webhook
```
timebertt, line 140: You could add the `sourceRepository` field

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -306,6 +319,12 @@ var (
 					{Type: &rbacv1.RoleBinding{}, Name: openstack.UsernamePrefix + openstack.CSIResizerName},
 				},
 			},
+			{
+				Name: stackit.PodIdentityWebhookName,
+				Objects: []*chart.Object{
+					{Type: &admissionregistrationv1.MutatingWebhookConfiguration{}, Name: stackit.PodIdentityWebhookName},
```
timebertt, line 327: Nit: this is missing the RBAC objects, but it shouldn't be relevant

timebertt, line 327: Nit: this is missing the RBAC objects, but it shouldn't be relevant

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+            - --cert-dir=/etc/webhook/certs
+            - --port={{ .Values.webhook.port }}
+          env:
+          - name: KUBECONFIG
+            value: /var/run/secrets/gardener.cloud/shoot/generic-kubeconfig/kubeconfig
+          ports:
+            - name: https
+              containerPort: {{ .Values.webhook.port }}
+              protocol: TCP
+            - name: metrics
```
timebertt, line 57: Should we collect metrics from this component?
See https://github.com/gardener/gardener/blob/master/docs/development/component-checklist.md#observability--operations-productivity

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: stackit-pod-identity-webhook
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: stackit-pod-identity-webhook
+    high-availability-config.resources.gardener.cloud/type: server
+spec:
```
timebertt, line 9: `revisionHistoryLimit` is missing, ref https://github.com/gardener/gardener/blob/master/docs/development/component-checklist.md#observability--operations-productivity

timebertt (dismissed):
Awesome, this looks great. Thanks for addressing my suggestions!

charts/internal/seed-controlplane/requirements.yaml
jastBytes: It has not real effect setting it via the `values.yaml` since it doesn't prevent rendering of resources in the templates. It only works as part of the `requirements.yaml`. IMHO having this in the values would lead to confusion since it wouldn't have any effect setting it to false.

charts/internal/shoot-system-components/requirements.yaml
jastBytes: It has not real effect setting it via the values.yaml since it doesn't prevent rendering of resources in the templates. It only works as part of the requirements.yaml. IMHO having this in the values would lead to confusion since it wouldn't have any effect setting it to false.

imagevector/images.yaml
jastBytes: Not sure. Maybe this is problematic. Will remove the digest.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
jastBytes: Wouldn't hurt I guess. I am not completely sure I understand the gardener docs. Do I need to add an https://github.com/gardener/gardener/blob/master/docs/extensions/logging-and-monitoring.md#servicemonitor or is there some magic that I am missing?

charts/internal/seed-controlplane/requirements.yaml
timebertt: But there are other similar values in `values.yaml`. Should we add this one or remove all of these?

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
timebertt: Probably, a `ServiceMonitor` is good enough for supporting Gardener's built-in monitoring setup.
Note that SKE disables Gardener's built-in monitoring and uses its own monitoring extension. For testing the `ServiceMonitor`, you would need to explicitly enable the built-in monitoring on an ondemand. For the SKE monitoring, you would need to add a corresponding scrape configuration to the monitoring extension.

charts/internal/seed-controlplane/requirements.yaml
jastBytes: I guess we could remove the others as well. Will check.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
jastBytes: Added the neccessary labels to the deployment to get scraped to the deployment. Will create a separate task for the scrape configuration in SKE monitoring.

charts/internal/seed-controlplane/requirements.yaml
jastBytes: I didn't find any place where this is part of the values where it is baiscally not doing anything. Instead I found myself missing places where I should've disabled it by default. Added those.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/vpa.yaml
```diff
@@ -14,8 +14,12 @@ spec:
...
     - containerName: stackit-pod-identity-webhook
       minAllowed:
-        memory: 80M
+        cpu: {{ .Values.vpa.resourcePolicy.minAllowed.cpu }}
+        memory: {{ .Values.vpa.resourcePolicy.minAllowed.memory }}
       maxAllowed:
         cpu: {{ .Values.vpa.resourcePolicy.maxAllowed.cpu }}
         memory: {{ .Values.vpa.resourcePolicy.maxAllowed.memory }}
-      controlledValues: RequestsOnly
+      controlledValues: RequestsAndLimits
```
timebertt: You don't set limits in the deployment 👀

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
```diff
@@ -20,6 +20,9 @@ spec:
         gardener.cloud/role: controlplane
         networking.gardener.cloud/to-dns: allowed
         networking.resources.gardener.cloud/to-kube-apiserver-tcp-443: allowed
+        prometheus.io/scrape: "true"
+        prometheus.io/port: {{ .Values.metrics.port | quote }}
+        prometheus.io/name: "{{ .Release.Name }}"
```
timebertt: I don't believe this works. Reading the code, it would need to be an annotation: https://github.com/gardener/gardener/blob/60a86001eca7145ff54b6555bcd17acea90d2933/pkg/component/observability/monitoring/prometheus/shoot/scrapeconfigs.go#L71-L138

I suggest making use of `ServiceMonitor` objects: https://github.com/gardener/gardener/blob/7c2403c7ed756245486d8ebbcdb6e17366f567ee/docs/extensions/logging-and-monitoring.md#servicemonitor

jastBytes: I guess the Gardener docs on that topic are a bit misleading. At least to me. Will have a look. Thank you!

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/vpa.yaml
jastBytes: Limits added.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
jastBytes: Added a service monitor. Will test in ondemand.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
```diff
@@ -72,6 +69,9 @@ spec:
               path: /readyz
               port: health
           resources:
+              limits:
+                cpu: {{ .Values.vpa.resourcePolicy.maxAllowed.cpu }}
+                memory: {{ .Values.vpa.resourcePolicy.maxAllowed.memory }}
```
timebertt: I'm confused now by all this resource and VPA stuff. Setting limits = maxAllowed means that limits will never be scaled, no? 
What's your goal here? Can you double-check other components deployed by Gardener for reference?

jastBytes: Sorry I did not quite get how `.resources` and the VPA play together. Setting limits and requests to the same values as for the minAllowed and maxAllowed settings of the VPA is quite wrong..

Gardener usually sets initial requests for its components to ensure they schedule correctly, but relies on the VPA's mutating webhook to right-size them on the fly. They tend not to use limits at all.

I will now remove the limits, set proper requests and let the VPA do the rest like Gardener approaches this in general.

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/deployment.yaml
jastBytes: Tested in ondemand, target is picked up by the shoot prometheus and is healty:

grafik

charts/internal/seed-controlplane/charts/stackit-pod-identity-webhook/templates/vpa.yaml
```diff
@@ -0,0 +1,25 @@
...
+  resourcePolicy:
+    containerPolicies:
+    - containerName: stackit-pod-identity-webhook
+      minAllowed:
+        cpu: {{ .Values.vpa.resourcePolicy.minAllowed.cpu }}
+        memory: {{ .Values.vpa.resourcePolicy.minAllowed.memory }}
+      maxAllowed:
+        cpu: {{ .Values.vpa.resourcePolicy.maxAllowed.cpu }}
+        memory: {{ .Values.vpa.resourcePolicy.maxAllowed.memory }}
+      controlledValues: RequestsAndLimits
```
timebertt: Final nit ;)
```suggestion
      controlledValues: RequestsOnly
```

timebertt (commented):
Awesome! Thanks for all your efforts and for working through my suggestions :) 
/lgtm

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -1276,3 +1301,45 @@ func cleanupCloudProviderConfigSecret(ctx context.Context, client k8sclient.Clie
...
+	secretsReader secretsmanager.Reader,
+	scaledDown bool,
+) (map[string]any, error) {
+	tlsSecret, found := secretsReader.Get(stackitPodIdentityWebhookServerName)
+	if !found {
+		return nil, fmt.Errorf("secret %q not found", stackitPodIdentityWebhookServerName)
+	}
+
+	return map[string]any{
+		"enabled":      feature.Gate.Enabled(feature.EnableSTACKITWorkloadIdentity),
```
dergeberl: Should this be always enabled for a shoot? This is only working when the `shoot.spec.kubernetes.kubeAPIServer.serviceAccountConfig.issuer` or the `authentication.gardener.cloud/issuer=managed` annotation is set. Does it make sense to check this config settings or add a possibility to enable/disable it for within the extension config? Keep in mind not only we use this extension.

dergeberl (changes requested):

pkg/controller/controlplane/valuesprovider.go
jastBytes: Good point! We could definitly check if the annotation is set additionally to the feature flag on the shoot. Would that be enough from your point of view?

dergeberl: It should be also possible to get it running with that setting of `shoot.spec.kubernetes.kubeAPIServer.serviceAccountConfig.issuer`? Enable it if one of the 2 options is set in addition to the feature gate sound like a good idea to me.

@MichaelEischer correct or do I miss something?

jastBytes: Sounds about right to me.

jastBytes: I've implemented the suggested changes. Could you have a look please? Thanks!

dergeberl (approved):
/lgtm

PR #54, opened 2026-03-17
Add application load balancer controller with ingress support

dergeberl:

/kind enhancement

This PR adds the application load balancer controller ([stackitcloud/application-load-balancer-controller](https://github.com/stackitcloud/application-load-balancer-controller)) to the gardener-extension-provider-stackit.

There was already some tests merged in this repo which did not work. So it should be safe to change that parts.

This PR removes the existing config field `deployALBIngressController`, removes the old feature gate, name things consistently, add a validation with config option if alb controller is allowed

This PR makes use of the validation of the shoot via the admission webhook. Therefore this PR is based on #203 and has to be merged after this.

/hold
The application load balancer controller is currently still under development and this PR uses a feature build (see PR https://github.com/stackitcloud/application-load-balancer-controller/pull/3). We only should merge this after we adopted the version in the imagevector to a released version. 

Fixes #

**Breaking changes**:


charts/gardener-extension-admission-stackit/charts/runtime/templates/deployment.yaml
```diff
@@ -63,6 +63,9 @@ spec:
         {{- if .Values.flatcarImageVersion }}
         - --flatcar-image-version={{ .Values.flatcarImageVersion }}
         {{- end }}
+        {{- if .Values.allowApplicationLoadBalancerController }}
+        - --allow-application-load-balancer-controller={{ .Values.allowApplicationLoadBalancerController }}
```
breuerfelix, line 67: I would call it `enable-application-load-balancer-controller`, but thats just personal preference
The ADR states `allow` but if thats still up for discussion, i prefer `enable` just optional

pkg/admission/validator/shoot.go
```diff
@@ -45,7 +49,7 @@ func (s *shoot) Validate(_ context.Context, newObj, oldObj client.Object) error
 
 	allErrs := field.ErrorList{}
 
-	allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfig(cpConfig, shoot.Spec.Kubernetes.Version, field.NewPath("spec").Child("provider").Child("controlPlaneConfig"))...)
+	allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfig(cpConfig, shoot.Spec.Kubernetes.Version, s.allowApplicationLoadBalancerController, field.NewPath("spec").Child("provider").Child("controlPlaneConfig").Child("controlPlaneConfig"))...)
```
breuerfelix:
```suggestion
	allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfig(cpConfig, shoot.Spec.Kubernetes.Version, s.allowApplicationLoadBalancerController, field.NewPath("spec").Child("provider").Child("controlPlaneConfig"))...)
```

pkg/apis/stackit/validation/controlplane_test.go
```diff
@@ -51,7 +51,7 @@ var _ = Describe("ControlPlaneConfig validation", func() {
 			controlPlane.Storage = &stackitv1alpha1.Storage{
 				CSI: &stackitv1alpha1.CSI{Name: "stackit", CompatibilityMode: "bogus"},
 			}
-			Expect(ValidateControlPlaneConfig(controlPlane, "", nilPath)).To(ConsistOf(
+			Expect(ValidateControlPlaneConfig(controlPlane, "", false, nilPath)).To(ConsistOf(
```
breuerfelix, line 99: lets create a test where allowALB is true that tests that at least one controller has to be enabled

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -1041,56 +1040,68 @@ func getCSIControllerChartValues(cluster *extensionscontroller.Cluster, userAgen
...
+	if apiEndpoints != nil && apiEndpoints.TokenEndpoint != nil {
+		values["tokenEndpoint"] = *apiEndpoints.TokenEndpoint
 	}
 
 	return values, nil
 }
 
-func DeploySTACKITALB(cpConfig *stackitv1alpha1.ControlPlaneConfig) bool {
+func DeploySTACKITApplicationLoadBalancer(cpConfig *stackitv1alpha1.ControlPlaneConfig) bool {
 	return ptr.Deref(cpConfig.ApplicationLoadBalancer, stackitv1alpha1.ApplicationLoadBalancerConfig{}).Enabled
```
breuerfelix, line 1092: we should add a check if the cli arg `allow-application....` is also true

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -205,12 +205,12 @@ var (
...
-				Images: []string{imagevector.ImageNameStackitAlbControllerManager},
+				Name:   openstack.STACKITApplicationLoadBalancerControllerName,
+				Images: []string{imagevector.ImageNameStackitApplicationLoadBalancerController},
 				Objects: []*chart.Object{
-					// stackit-alb-controller-manager
-					{Type: &appsv1.Deployment{}, Name: openstack.STACKITALBControllerManagerName},
-					{Type: &vpaautoscalingv1.VerticalPodAutoscaler{}, Name: openstack.STACKITALBControllerManagerName},
+					// stackit-application-load-balancer-controller
+					{Type: &appsv1.Deployment{}, Name: openstack.STACKITApplicationLoadBalancerControllerName},
+					{Type: &vpaautoscalingv1.VerticalPodAutoscaler{}, Name: openstack.STACKITApplicationLoadBalancerControllerName},
```
breuerfelix, line 211: you need to add the pdb and configmap here aswell, otherwise they will get orphaned

charts/internal/seed-controlplane/charts/stackit-application-load-balancer-controller/values.yaml
```diff
@@ -0,0 +1,21 @@
...
+    #  applicationLoadBalancerApi: "foo"
+    #  applicationLoadBalancerCertificateApi: "foo"
+  applicationLoadBalancer:
+    networkId: ""
+#tokenEndpoint: "foo"
+metricsPort: 9090
+podAnnotations: {}
+podLabels: {}
+images:
+  stackit-alb-controller-manager: image:tag
```
breuerfelix:
```suggestion
  stackit-application-load-balancer-controller: image:tag
```

pkg/controller/controlplane/valuesprovider.go
breuerfelix: don't we also have to check if ingress is enabled?
does it make sense to just import the function we use in the validating webhook?

or do we intentionally keep the validation in the webhook and just do not care about double checking here... i guess i am fine with both

dergeberl: The cli flag is on a different deployment, which is deployed with a different helmchart. So I dont have the details what is set here.

We can check for the Ingress here but the admission should only allow with ingress enabled.

charts/gardener-extension-admission-stackit/charts/runtime/templates/deployment.yaml
dergeberl: I prefer `allow`. As this field does not enable the controller it is allowing the customer/user to enable it in the shoot.

pkg/controller/controlplane/valuesprovider.go
breuerfelix: Yes i don't have a strong opinion. One one side i like that the validation is done where validation should happen and not everything is double checked here, but on the other side the "correct" way would be to also check it here.

lets keep it like that.

imagevector/images.yaml
```diff
@@ -133,9 +133,11 @@ images:
   repository: registry.k8s.io/sig-storage/livenessprobe
   tag: "v2.18.0"
 
-- name: stackit-alb-controller-manager
-  repository: reg3.infra.ske.eu01.stackit.cloud/temp/alb-controller-manager
-  tag: "1245"
+- name: stackit-application-load-balancer-controller
+  sourceRepository: github.com/stackitcloud/application-load-balancer-controller
+  repository: ghcr.io/stackitcloud/application-load-balancer-controller-dev
+  tag: "albc"
```
breuerfelix: just as a reminder:
we have to use a proper image + version here

breuerfelix:
only the prefix override is failing

breuerfelix (approved):

nschad (approved):

nschad:

PR #103, opened 2026-04-15
[GEP-36] Implement `SelfHostedShootExposure` controller using STACKIT NLB

jamand:

/kind enhancement

As part of the Gardener [GEP-36: Self-Hosted Shoot Exposure](https://github.com/gardener/enhancements/tree/main/geps/0036-self-hosted-shoot-exposure), this PR adds a controller for the `SelfHostedShootExposure` resource in the extension.

The controller fulfils the respective [`SelfHostedShootExposure` contract](https://github.com/gardener/gardener/blob/master/docs/extensions/resources/selfhostedshootexposure.md) and Extension Controller Interface described in the GEP. It used the STACKIT Network Loadbalancing service as backing service to expose the endpoints referenced in the resource Spec, which can be updated by the Gardenlet.

Part of https://github.com/gardener/gardener/issues/13602

I tested this using two approaches:
1) Manual testing via exposing a `DaemonSet` `nginx` on a worker pool and providing the worker IPs in a `SelfHostedShootExposure` picked up by the Extension, you may use `make mirrord-run` for that.
2) Running integration tests I added (which test basic operations like creation, update of target pools, plan, deletion).

The `gardener-extension-provider-stackit/test/integration/selfhostedshootexposure/stackit` is not registered as a prow check currently, I could add it if it's worth it (the tests take about 3 minutes or so).

/cc @stackitcloud/ske-gardener


jamand:
Ah, I have some questions that came to my mind where I'm not sure what if I can completely handle it here or if we need a follow-up? @timebertt 
As I've just finished working on https://github.com/gardener/gardener/pull/14420 I realised that the control-plane behind self-hosted shoot exposure does not have any access control implemented now. It's not directly part of the spec. I could add it into the Provider-specific specs and do ACLs at the load balancer level. Then we wouldn't need to care about PROXY Protocol and needing to pass the Client IP through. As it's just the control plane and not different endpoints (like discovery server, ske-api) that may require different access levels, there is no hard need for that to my understanding. We just have to check how to use existing logic to update the ACLs for dynamically changing CIDRs like we do in other use cases.

jamand:
I added a way to provide `AllowedSourceRanges` in the Provider specs for now (also setting, modifying, removal tested in the integration tests) in the last commit.

test/integration/selfhostedshootexposure/selfhostedshootexposure_suite_test.go
timebertt, line 1: > The gardener-extension-provider-stackit/test/integration/selfhostedshootexposure/stackit is not registered as a prow check currently, I could add it if it's worth it (the tests take about 3 minutes or so).

If we add the test code here, we should also run it in CI to ensure the tests keep working.

pkg/controller/selfhostedshootexposure/actuator.go
```diff
@@ -0,0 +1,126 @@
...
+	// For now, we only support secret-based credentials via ObjectReference pointing to a Secret.
+	var secretRef corev1.SecretReference
+	if exposure.Spec.CredentialsRef != nil {
+		// Use the explicitly provided ObjectReference (must point to a Secret for now)
+		secretRef = corev1.SecretReference{
+			Name:      exposure.Spec.CredentialsRef.Name,
+			Namespace: exposure.Spec.CredentialsRef.Namespace,
+		}
+	} else {
+		// Fall back to the default cloud-provider secret
```
timebertt: We don't need this fallback. If the shoot has managed infrastructure, Gardener will publish both the `CredentialsRef` and the `cloudprovider` secret. Otherwise, it won't publish either.
I.e., you can blindly pass the values from `exposure.Spec.CredentialsRef` to the client constructor.

pkg/stackit/client/loadbalancing.go
```diff
@@ -11,9 +11,14 @@ import (
 )
 
 type LoadBalancingClient interface {
+	ProjectID() string
+
 	ListLoadBalancers(ctx context.Context) ([]loadbalancer.LoadBalancer, error)
 	DeleteLoadBalancer(ctx context.Context, lbName string) error
```
timebertt: Nit: can you sort the methods in the "usual order" like in the iaas client?

pkg/stackit/client/loadbalancing.go
```diff
@@ -11,9 +11,14 @@ import (
...
 	GetLoadBalancer(ctx context.Context, id string) (*loadbalancer.LoadBalancer, error)
+	CreateLoadBalancer(ctx context.Context, payload loadbalancer.CreateLoadBalancerPayload) (*loadbalancer.LoadBalancer, error)
+	UpdateLoadBalancer(ctx context.Context, lbName string, payload loadbalancer.UpdateLoadBalancerPayload) (*loadbalancer.LoadBalancer, error)
+	UpdateLoadBalancerTargetPool(ctx context.Context, lbName, tpName string, payload loadbalancer.UpdateTargetPoolPayload) (*loadbalancer.TargetPool, error)
```
timebertt: Try to avoid abbreviations. I would replace `lbName` with `name`/`loadBalancerName` and `tpName` with `targetPool`/`targetPoolName`

pkg/stackit/client/loadbalancing.go
```diff
@@ -54,5 +63,24 @@ func (l loadBalancingClient) DeleteLoadBalancer(ctx context.Context, lbName stri
 }
 
 func (l loadBalancingClient) GetLoadBalancer(ctx context.Context, lbName string) (*loadbalancer.LoadBalancer, error) {
-	return l.Client.GetLoadBalancer(ctx, l.projectID, l.region, lbName).Execute()
+	lb, err := l.Client.GetLoadBalancer(ctx, l.projectID, l.region, lbName).Execute()
+	if err != nil {
+		if IsNotFound(err) {
+			return nil, nil
```
timebertt: Code calling the client might want to (and should) handle a `NotFound` error explicitly. Let's remove the handling here to allow that.

charts/gardener-extension-provider-stackit/templates/deployment.yaml
```diff
@@ -66,6 +66,7 @@ spec:
         - --heartbeat-namespace={{ .Release.Namespace }}
         - --heartbeat-renew-interval-seconds={{ .Values.controllers.heartbeat.renewIntervalSeconds }}
         - --infrastructure-max-concurrent-reconciles={{ .Values.controllers.infrastructure.concurrentSyncs }}
+        - --selfhostedshootexposure-max-concurrent-reconciles={{ .Values.controllers.selfhostedshootexposure.concurrentSyncs }}
         - --ignore-operation-annotation={{ .Values.controllers.ignoreOperationAnnotation }}
```
timebertt, line 69: This is somewhat sorted alphabetically. Can you move it down one line?

pkg/apis/stackit/v1alpha1/register.go
```diff
@@ -45,6 +45,7 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&ControlPlaneConfig{},
 		&WorkerStatus{},
 		&WorkerConfig{},
+		&SelfHostedShootExposureConfig{},
 	)
```
timebertt, line 49: Sort alphabetically?

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+				opts.PlanId = *providerConfig.LoadBalancer.PlanId
+			}
+			if providerConfig.LoadBalancer.AccessControl != nil {
+				opts.AllowedSourceRanges = providerConfig.LoadBalancer.AccessControl.AllowedSourceRanges
+			}
+		}
+	}
+	// Default plan if not specified
+	if opts.PlanId == "" {
+		opts.PlanId = "p10"
```
timebertt: Instead of defaulting this in the controller code, I suggest implementing proper API defaulting in `pkg/aips/stackit/v1alpha1`. You can check https://github.com/gardener/gardener/blob/7f3403e036962a7d3be93f2ad07bce330ca040a7/pkg/apis/core/v1beta1/defaults_project.go for an example.

pkg/apis/stackit/v1alpha1/types_selfhostedshootexposure.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// LoadBalancer contains configuration for the load balancer.
+	// +optional
+	LoadBalancer *LoadBalancerConfig `json:"loadBalancer,omitempty"`
+}
+
+// LoadBalancerConfig contains configuration for the load balancer.
+type LoadBalancerConfig struct {
+	// PlanId specifies the service plan (size) of the load balancer.
+	// Currently supported plans are p10, p50, p250, p750 (compare API docs).
```
timebertt, line 22:
```suggestion
	// See https://docs.stackit.cloud/products/network/load-balancing-and-content-delivery/network-load-balancer/reference/service-plans/
	// Defaults to "p10".
```

pkg/apis/stackit/v1alpha1/types_selfhostedshootexposure.go
```diff
@@ -0,0 +1,25 @@
...
+	// +optional
+	PlanId *string `json:"planId,omitempty"`
```
timebertt:
```suggestion
	PlanID *string `json:"planID,omitempty"`
```

pkg/apis/stackit/v1alpha1/types_selfhostedshootexposure.go
```diff
@@ -0,0 +1,36 @@
...
+// SelfHostedShootExposureConfig contains configuration settings for exposing self-hosted shoots.
+type SelfHostedShootExposureConfig struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// LoadBalancer contains configuration for the load balancer.
+	// +optional
+	LoadBalancer *LoadBalancerConfig `json:"loadBalancer,omitempty"`
+}
+
+// LoadBalancerConfig contains configuration for the load balancer.
```
timebertt: Nit: you could remove the `Config` suffix from `LoadBalancerConfig` and `AccessControlConfig` to make it a bit more comfortable to type.

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (a *Actuator) DetermineOptions(ctx context.Context, exposure *extensionsv1alpha1.SelfHostedShootExposure, cluster *extensionscontroller.Cluster, projectID string) (*Options, error) {
+	opts := &Options{
+		SelfHostedShootExposure: exposure,
+		ProjectID:               projectID,
+		ResourceName:            fmt.Sprintf("%s-exposure-%s", cluster.Shoot.Status.TechnicalID, exposure.Name),
+		// STACKIT LB labels do not allow '/' in keys, so we use the flat dot-separated form
+		// matching the convention used for other STACKIT LBs (see controlplane.STACKITLBClusterLabelKey).
+		Labels: map[string]string{
```
timebertt, line 67: Can you respect the `CustomLabelDomain` config in this controller as well?

pkg/controller/selfhostedshootexposure/actuator.go
```diff
@@ -0,0 +1,126 @@
...
+	return ingresses, util.DetermineError(err, helper.KnownCodes)
+}
+
+func (a *Actuator) reconcile(ctx context.Context, log logr.Logger, exposure *extensionsv1alpha1.SelfHostedShootExposure, cluster *extensionscontroller.Cluster) ([]corev1.LoadBalancerIngress, error) {
+	r, err := a.getResources(ctx, log, exposure, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := r.reconcileLoadBalancer(ctx, log); err != nil {
```
timebertt, line 49: I would expect that we also need to create an additional security group rule in the cluster's security group to allow traffic on port `443` (`SelfHostedShootExposure.spec.port`) which is outside of the already allowed node port range (`30000-32767`)

pkg/controller/selfhostedshootexposure/actuator.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if err := r.deleteLoadBalancer(ctx, log); err != nil {
+		return fmt.Errorf("error deleting load balancer: %w", err)
+	}
+
+	return nil
+}
+
+func (a *Actuator) ForceDelete(ctx context.Context, log logr.Logger, exposure *extensionsv1alpha1.SelfHostedShootExposure, cluster *extensionscontroller.Cluster) error {
+	return a.Delete(ctx, log, exposure, cluster)
```
timebertt: Shouldn't we do nothing here? (see Bastion controller)
E.g., this call should not return an error if the credentials are invalid.
See https://github.com/gardener/gardener/blob/master/docs/extensions/force-deletion.md

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+	ResourceName string
+	// Labels added to all STACKIT resources.
+	Labels map[string]string
+
+	// Region for the LB, determined from Cluster.spec.shoot.spec.region (RegionOne is replaced with eu01).
+	Region string
+	// NetworkID is the ID of the network where the control plane nodes reside.
+	NetworkID string
+	// PlanId specifies the service plan (size) of the load balancer.
+	PlanId string
```
timebertt: `PlanID`

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+	// AllowedSourceRanges restricts which source CIDRs may reach the load balancer.
+	// Empty means unrestricted.
+	AllowedSourceRanges []string
+}
+
+func (a *Actuator) DetermineOptions(ctx context.Context, exposure *extensionsv1alpha1.SelfHostedShootExposure, cluster *extensionscontroller.Cluster, projectID string) (*Options, error) {
+	opts := &Options{
+		SelfHostedShootExposure: exposure,
+		ProjectID:               projectID,
+		ResourceName:            fmt.Sprintf("%s-exposure-%s", cluster.Shoot.Status.TechnicalID, exposure.Name),
```
timebertt: I think you can omit the `exposure.Name` part, because it will be static/meaningless anyway.

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	opts.NetworkID = infraStatus.Networks.ID
+
+	// Decode providerConfig to extract STACKIT-specific settings
+	if exposure.Spec.ProviderConfig != nil {
+		providerConfig := &stackitv1alpha1.SelfHostedShootExposureConfig{}
+		if _, _, err := a.Decoder.Decode(exposure.Spec.ProviderConfig.Raw, nil, providerConfig); err != nil {
+			return nil, fmt.Errorf("error decoding providerConfig: %w", err)
+		}
+		if errs := validation.ValidateSelfHostedShootExposureConfig(providerConfig, field.NewPath("providerConfig")); len(errs) > 0 {
```
timebertt:
```suggestion
		if errs := validation.ValidateSelfHostedShootExposureConfig(providerConfig, field.NewPath("spec.providerConfig")); len(errs) > 0 {
```

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return opts, nil
+}
+
+func getInfrastructureStatus(ctx context.Context, c client.Client, cluster *extensionscontroller.Cluster) (*stackitv1alpha1.InfrastructureStatus, error) {
+	infra := &extensionsv1alpha1.Infrastructure{}
+	if err := c.Get(ctx, client.ObjectKey{Namespace: cluster.ObjectMeta.Name, Name: cluster.Shoot.Name}, infra); err != nil {
+		if apierrors.IsNotFound(err) {
+			// Infrastructure is reconciled before SelfHostedShootExposure; absence is a normal transient state on initial creation.
```
timebertt: Why is that?

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,116 @@
...
+			return nil, &reconcilerutils.RequeueAfterError{
+				RequeueAfter: 30 * time.Second,
+				Cause:        fmt.Errorf("waiting for Infrastructure resource to be created"),
+			}
+		}
+		return nil, fmt.Errorf("error getting infrastructure: %w", err)
+	}
+	if infra.Status.ProviderStatus == nil {
```
timebertt: Isn't this handled by `helper.InfrastructureStatusFromRaw` already?

pkg/controller/selfhostedshootexposure/resources.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (r *Resources) getExistingResources(ctx context.Context, log logr.Logger) error {
+	lb, err := r.LBClient.GetLoadBalancer(ctx, r.ResourceName)
+	if err != nil {
+		return fmt.Errorf("error getting load balancer: %w", err)
+	}
+	if lb != nil {
+		r.LoadBalancer = lb
+		log.V(1).Info("Found existing load balancer", "loadbalancer", r.LoadBalancer.GetName())
```
timebertt:
```suggestion
		log.V(1).Info("Found existing load balancer", "loadBalancer", r.LoadBalancer.GetName())
```

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+// deprecated top-level stackit-sdk-go/services/loadbalancer package.
+const (
+	// lbErrTypeTargetNotActive encodes that target may not be ready (yet).
+	lbErrTypeTargetNotActive = "TYPE_TARGET_NOT_ACTIVE"
+)
+
+func (r *Resources) reconcileLoadBalancer(ctx context.Context, log logr.Logger) error {
+	targets, err := r.buildTargets()
+	if err != nil {
+		return err
```
timebertt: Wrapping the error in places like this helps identify where an error comes from easily.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		}
+		targets[i] = loadbalancer.Target{
+			DisplayName: &endpoint.NodeName,
+			Ip:          &ip,
+		}
+	}
+
+	// Sort targets by IP (primary) and DisplayName (secondary) for deterministic ordering
+	sort.Slice(targets, func(i, j int) bool {
+		if *targets[i].Ip != *targets[j].Ip {
```
timebertt: You can simplify this to
```go
		return cmp.Or(
			cmp.Less(*targets[i].Ip, *targets[j].Ip),
			cmp.Less(*targets[i].DisplayName, *targets[j].DisplayName),
		)
```

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		return r.updateTargetPool(ctx, log, targets)
+	}
+
+	// Full-state update: STACKIT's UpdateLoadBalancer (PUT endpoint)
+	return r.updateLoadBalancer(ctx, log, targets)
+}
+
+func (r *Resources) createLoadBalancer(ctx context.Context, log logr.Logger, targets []loadbalancer.Target) error {
+	if len(targets) == 0 {
+		// Endpoints are populated asynchronously by gardenlet from healthy control-plane nodes.
```
timebertt: Actually, `spec.endpoints` should never be empty. gardenadm will populate it right away

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		// Empty endpoints on first create is a normal transient state.
+		return &reconcilerutils.RequeueAfterError{
+			RequeueAfter: 30 * time.Second,
+			Cause:        fmt.Errorf("waiting for endpoints to be populated in spec"),
+		}
+	}
+
+	log.V(1).Info("Creating load balancer", "loadBalancer", r.ResourceName, "networkID", r.NetworkID, "planID", r.PlanId)
```
timebertt: Consider dropping this log. If the call succeeds, the creation will be logged, otherwise, the error will be logged.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	loadbalancer "github.com/stackitcloud/stackit-sdk-go/services/loadbalancer/v2api"
+	loadbalancerwait "github.com/stackitcloud/stackit-sdk-go/services/loadbalancer/v2api/wait"
+	corev1 "k8s.io/api/core/v1"
+
+	stackitclient "github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/stackit/client"
+)
+
+const (
+	// lbNetworkRoleListenersAndTargets is the network role for listeners and targets in the load balancer network.
+	lbNetworkRoleListenersAndTargets = "ROLE_LISTENERS_AND_TARGETS"
```
timebertt: Can we use constants from the SDK instead? E.g., `loadbalancer.NETWORKROLE_LISTENERS_AND_TARGETS` and `loadbalancer.LISTENERPROTOCOL_TCP`?

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	// protocolTCP is the TCP protocol identifier for the listener.
+	protocolTCP = "PROTOCOL_TCP"
+	// listenerName is the (single) hardcoded listener name for exposing the control plane API server.
+	listenerName = "listener-control-plane"
+	// targetPoolName is the (single) hardcoded target pool name for control plane nodes.
+	targetPoolName = "target-pool-control-plane"
```
timebertt: Consider dropping the `listener-` and `target-pool-` prefixes as they are implied by the resource/context

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+)
+
+// STACKIT LB error types reported in LoadBalancer.Errors[].Type. v2api weakened Type from a
+// generated enum to *string (known openapi-generator limitation confirmed with the LB team);
+// the authoritative set of values still lives as LOADBALANCERERRORTYPE_* constants in the
+// deprecated top-level stackit-sdk-go/services/loadbalancer package.
+const (
```
timebertt: Nit: `const` blocks can be merged

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	return slices.Compact(slices.Sorted(slices.Values(s)))
+}
+
+// wrapLBAPIError classifies STACKIT LB API errors: 409 Conflicts are transient (another caller
+// modified the LB between our GET and our write) and are retried via RequeueAfterError; anything
+// else is returned as a regular error so Gardener can classify + surface it.
+func wrapLBAPIError(op string, err error) error {
+	if stackitclient.IsConflict(err) {
+		return &reconcilerutils.RequeueAfterError{
+			RequeueAfter: 15 * time.Second,
```
timebertt: Typically, controllers handle conflict errors with exponential backoff with jittering (default behavior when returning an error). When using periodic retries like this one, the behavior is less likely to converge.
Apart from that, we would like to reduce the rate of API calls after a bit in case, the errors keep happening.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+	return true
+}
+
+// targetPoolNeedsUpdate checks if the target pool in the load balancer needs updating.
+// specTargets should be pre-built to avoid double-building.
+func (r *Resources) targetPoolNeedsUpdate(specTargets []loadbalancer.Target) (bool, error) {
+	// If no load balancer exists yet, no update needed (will be created fresh)
+	if r.LoadBalancer == nil {
+		return false, nil
```
timebertt: This is only called by `reconcileLoadBalancer` if the LB already exists. Probably returning an error here is better

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+
+	// If LB exists but has no target pools, check if spec has targets
+	if len(r.LoadBalancer.TargetPools) == 0 {
+		return len(specTargets) > 0, nil
+	}
+
+	// Validate that the load balancer has the expected target pool
+	if r.LoadBalancer.TargetPools[0].Name == nil || *r.LoadBalancer.TargetPools[0].Name != targetPoolName {
```
timebertt: You can use `ptr.Deref` as a shortcut for this

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		actualName := ""
+		if r.LoadBalancer.TargetPools[0].Name != nil {
+			actualName = *r.LoadBalancer.TargetPools[0].Name
+		}
+		return false, fmt.Errorf("unexpected target pool name: expected %q, got %q",
+			targetPoolName, actualName)
```
timebertt: Shouldn't we update the LB with the new name instead of failing here?

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	// If LB exists but has no target pools, check if spec has targets
+	if len(r.LoadBalancer.TargetPools) == 0 {
+		return len(specTargets) > 0, nil
+	}
+
+	// Validate that the load balancer has the expected target pool
+	if r.LoadBalancer.TargetPools[0].Name == nil || *r.LoadBalancer.TargetPools[0].Name != targetPoolName {
+		actualName := ""
+		if r.LoadBalancer.TargetPools[0].Name != nil {
+			actualName = *r.LoadBalancer.TargetPools[0].Name
```
timebertt: Same here, use `ptr.Deref`

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+			targetPoolName, actualName)
+	}
+
+	// Get targets from the first target pool and copy before sorting
+	lbTargets := make([]loadbalancer.Target, len(r.LoadBalancer.TargetPools[0].Targets))
+	copy(lbTargets, r.LoadBalancer.TargetPools[0].Targets)
+
+	// Sort the LB targets for comparison (same order as spec targets)
+	sort.Slice(lbTargets, func(i, j int) bool {
+		if *lbTargets[i].Ip != *lbTargets[j].Ip {
```
timebertt: Instead of duplicating the logic here, consider extracting a `sortTargets` function

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		if addr.Type == corev1.NodeInternalIP {
+			return addr.Address, nil
+		}
+	}
+	return "", fmt.Errorf("endpoint %s has no InternalIP address", endpoint.NodeName)
+}
+
+// targetsEqual compares two target lists for equality.
+// Both lists should be sorted by IP for correct comparison.
+func targetsEqual(spec, lb []loadbalancer.Target) bool {
```
timebertt: Consider using `cmp.Equal` instead of manually implementing this function. See https://github.com/stackitcloud/gardener-extension-provider-stackit/blob/51182412f829a15b9f462731b70df7753ff0d278/pkg/stackit/client/iaas.go#L253 for an example.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		current = r.LoadBalancer.Options.AccessControl.AllowedSourceRanges
+	}
+	return !stringSetsEqual(current, r.AllowedSourceRanges)
+}
+
+// stringSetsEqual compares two string slices as unordered sets, ignoring duplicates.
+// The LB API may or may not de-duplicate (causing reconciliation loop), remove duplicates
+// for a clean approach.
+func stringSetsEqual(a, b []string) bool {
+	return slices.Equal(sortedUnique(a), sortedUnique(b))
```
timebertt: I expect this to be more efficient:
````go
return sets.New(a...).Equal(sets.New(b...))
````

File: pkg/controller/selfhostedshootexposure/resources_loadbalancer.go [hunk 35]
```diff
@@ -0,0 +1,440 @@
...
+}
+
+func (r *Resources) updateTargetPool(ctx context.Context, log logr.Logger, targets []loadbalancer.Target) error {
+	log.Info("Target pool needs updating", "loadBalancer", r.ResourceName)
+	_, err := r.LBClient.UpdateLoadBalancerTargetPool(ctx,
+		r.ResourceName,
+		targetPoolName,
+		loadbalancer.UpdateTargetPoolPayload{
+			Name:       new(targetPoolName),
+			TargetPort: &r.SelfHostedShootExposure.Spec.Port,
```
timebertt: As a general rule, avoid sharing memory with k8s API objects. I.e., instead of referencing (`&`), create a new value on the stack and reference it (`new(...)` or `ptr.To(...)`).

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+		})
+	if err != nil {
+		return wrapLBAPIError("updating load balancer target pool", err)
+	}
+	log.Info("Updated load balancer target pool", "loadBalancer", r.ResourceName)
+
+	// Re-read the LB so downstream readiness checks don't see the pre-write status (STACKIT
+	// transitions the LB to STATUS_PENDING on any write). UpdateLoadBalancerTargetPool only
+	// returns the TargetPool, so a full GET is needed.
+	refreshed, err := r.LBClient.GetLoadBalancer(ctx, r.ResourceName)
```
timebertt: Reading this, I'm asking myself whether using the `UpdateLoadBalancerTargetPool` is actually the "fast path". Using the `UpdateLoadBalancer` method spares a second API call, right?

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,440 @@
...
+	if err != nil {
+		return fmt.Errorf("error refreshing load balancer after target pool update: %w", err)
+	}
+	r.LoadBalancer = refreshed
```
timebertt: Consider assigning to `r.LoadBalancer` directly instead of saving in an intermediate variable.

timebertt (changes requested):
Thanks for this PR!
Here is the first part of my review. I haven't quite made it through. E.g., I haven't reviewed the tests yet.

pkg/controller/selfhostedshootexposure/resources_loadbalancer_test.go
```diff
@@ -0,0 +1,795 @@
...
+		It("should requeue creation without endpoints", func() {
+			r.SelfHostedShootExposure.Spec.Endpoints = []extensionsv1alpha1.ControlPlaneEndpoint{}
+
+			err := r.reconcileLoadBalancer(ctx, logger)
+
+			Expect(err).To(HaveOccurred())
+			// Endpoints are populated asynchronously by gardenlet; empty endpoints should trigger a clean requeue,
+			// not a fatal error.
+			var rae *reconcilerutils.RequeueAfterError
+			Expect(errors.As(err, &rae)).To(BeTrue())
```
timebertt: Nit: you could use `errors.Is` here

pkg/controller/selfhostedshootexposure/resources_loadbalancer_test.go
```diff
@@ -0,0 +1,795 @@
...
+			Expect(rae.Cause.Error()).To(ContainSubstring("waiting for endpoints to be populated"))
```
timebertt: Nit: you could use `MatchError()` here

test/integration/testdata/upstream-crds/10-crd-extensions.gardener.cloud_selfhostedshootexposures.yaml
```diff
@@ -0,0 +1,396 @@
+---
```
timebertt, line 1: You should add this file to `make gardener-crds`

Makefile
```diff
@@ -173,6 +173,16 @@ test-integration-infra: $(REPORT_COLLECTOR) $(SETUP_ENVTEST) $(GINKGO) ## Run in
 		-- \
 		$(INFRA_TEST_FLAGS)
 
+.PHONY: test-integration-exposure
+test-integration-exposure: $(REPORT_COLLECTOR) $(SETUP_ENVTEST) $(GINKGO) ## Run selfhostedshootexposure integration tests
```
timebertt, line 186: Should be documented in `testing.md` as well

test/integration/selfhostedshootexposure/stackit/selfhostedshootexposure_test.go
```diff
@@ -0,0 +1,707 @@
...
+	}
+
+	restConfig, err := testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(restConfig).ToNot(BeNil())
+
+	httpClient, err := rest.HTTPClientFor(restConfig)
+	Expect(err).NotTo(HaveOccurred())
+	mapper, err := apiutil.NewDynamicRESTMapper(restConfig, httpClient)
+	Expect(err).NotTo(HaveOccurred())
```
timebertt: Setting up a dedicated/custom http client and rest mapper seems unnecessary

pkg/controller/selfhostedshootexposure/options.go
timebertt: On a second thought, keeping this `exposure.Name` part would ensure we don't need any migrations or compatibility code when Gardener later on finds another use case for creating multiple `SelfHostedShootExposure` objects.

test/integration/selfhostedshootexposure/stackit/selfhostedshootexposure_test.go
```diff
@@ -0,0 +1,707 @@
...
+		suffix, err := gardenerutils.GenerateRandomStringFromCharset(5, "0123456789abcdefghijklmnopqrstuvwxyz")
+		Expect(err).NotTo(HaveOccurred())
+		namespaceName = "stackit--exp-it--" + suffix
+		// ResourceName = "<technicalID>-exposure-<exposureName>"
+		lbName = namespaceName + "-exposure-" + exposureName
+	})
+
+	AfterEach(func() {
+		// Best-effort cleanup of LB via STACKIT API (in case the controller didn't delete it)
+		if lbName != "" {
```
timebertt: Can `lbName` ever be empty?

test/integration/selfhostedshootexposure/stackit/selfhostedshootexposure_test.go
```diff
@@ -0,0 +1,707 @@
...
+			lb, _ := lbClient.GetLoadBalancer(ctx, lbName)
```
timebertt: What about transient errors calling the API? 
You should probably wrap such calls in test in an `Eventually`

test/integration/selfhostedshootexposure/stackit/selfhostedshootexposure_test.go
```diff
@@ -0,0 +1,707 @@
...
+			if lb != nil {
+				log.Info("Cleaning up leftover load balancer", "name", lbName)
+				_ = lbClient.DeleteLoadBalancer(ctx, lbName)
```
timebertt: Same here, use `Eventually` to tolerate transient errors in API calls and wait for the load balancer to be deleted

test/integration/selfhostedshootexposure/stackit/selfhostedshootexposure_test.go
```diff
@@ -0,0 +1,707 @@
...
+				g.Expect(lb.Errors).NotTo(BeEmpty())
+				for _, e := range lb.Errors {
+					g.Expect(e.Type).NotTo(BeNil())
+					g.Expect(*e.Type).To(Equal("TYPE_TARGET_NOT_ACTIVE"))
+				}
+			}
+		}).WithTimeout(5 * time.Minute).WithPolling(5 * time.Second).Should(Succeed())
+
+		By("verify SelfHostedShootExposure CR is being reconciled")
+		Expect(c.Get(ctx, client.ObjectKeyFromObject(exposure), exposure)).To(Succeed())
```
timebertt: Use `Eventually` for waiting for controller actions on API object

timebertt (commented):
Made it through the PR. Let me know if you want to discuss any comments in a chat :)

pkg/controller/selfhostedshootexposure/actuator.go
jamand: Adapted. I also was not sure how the `WorkloadIdentity` flow can be integrated here, but I think I got it from the STACKIT SDK Readme now and will handle it.

pkg/stackit/client/loadbalancing.go
jamand: I'm not sure I get the order. After `ProjectID`, sometimes it's `Create -> Get -> Delete` (Network), then it's `Create -> Delete -> Get` (Security Group), then `Get -> Create -> Delete` (Keypair)?

For now I did CRUD (Create, Get, List, Update, Delete).

pkg/apis/stackit/v1alpha1/types_selfhostedshootexposure.go
jamand: The [Loadbalancer API specs](https://docs.api.eu01.stackit.cloud/documentation/load-balancer/version/v2#tag/Load-Balancer/operation/APIService_CreateLoadBalancer) say `PlanId` (which the STACKIT sdk then provide us as such) but I agree, `PlanID` is better and I'll use that here.

pkg/controller/selfhostedshootexposure/options.go
jamand: I removed it because the LBs are not able to handle regular `/` like other resources (e.g. IaaS), so you would need a different labeling than for all other resources (not using `BuildLabelKey`). Do you have a suggestion, e.g. using one of `[._-]` instead, so that it's for example `ske.stackit.cloud.cluster` instead of `ske.stackit.cloud/cluster`?

Also see: https://github.com/stackitcloud/gardener-extension-provider-stackit/blob/2d81ee245f55c6885e0c880222ad356a73a3c4c6/pkg/controller/infrastructure/openstack/infraflow/delete.go#L248-L249

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
jamand: I did an own constants here because there is a known issue with the OpenAPI spec generation that occurred when the versioning was introduced. The constant enums are defined in the deprecated `stackit-sdk-go/services/loadbalancer` package (e.g. `ListenerProtocol`), in the `v2api` package the type is now barely a `*string`. After consultation with Loadbalancing/STACKIT SDK we will use the types from the old package until there is a solution (might have to annotate the use to make the linter happy).

pkg/controller/selfhostedshootexposure/actuator.go
jamand: Discussed with @timebertt . This is up for later, as in this context `Gardener WorkloadIdentity != STACKIT WorkloadIdentity`. The whole topic to integrate that into our controllers is separate from this. For now, as we control creation of `SelfHostedShootExposure` CR in our environment, we are fine only supporting `CredentialsRefs` to `Secrets`.

pkg/controller/selfhostedshootexposure/actuator.go
jamand: From my understanding everything should be fine. When creating a `NLB`, two security groups, `backend-port` and `backend` are created. `backend` contains a rule that explicitly allows the `Spec.Port` set in the LB (with the protocol TCP we configured). The SecurityGroup is added to the machines targeted by the LB backend, and allows access to the port (although the regular SG on the machines would otherwise just allow the `NodePort` ranges).

I could verify this by creating a `DaemonSet` on worker nodes (shoot `s-eu01-000`) and referencing them in a `SelfHostedShootExposure` (created in shoot `r-eu01-0`). From my local machine I could successfully curl the endpoints via the LB.

test/integration/selfhostedshootexposure/selfhostedshootexposure_suite_test.go
jamand: Done in https://github.com/stackitcloud/ske-ci-infra/pull/4666

pkg/controller/selfhostedshootexposure/options.go
jamand: Checked with Infra that this aligns with the other controllers and CCM, we will add a similar TODO and as soon as LB supports this we will change it.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
jamand: You're right, I changed the handling for those helpers with the contract of the LB already existing.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
jamand: Ok, we do a PUT anyways, so we will just replace the slice of target pools by our expected single target pool with the constant name.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
jamand: Yes, you're right. This is a bit sad but I guess the API design makes this necessary (and it does not look like using the "normal" `UpdateLoadBalancer` method is slower (performance and rate limiting would be two things I would assume why `UpdateLoadBalancerTargetPool` exists).

jamand:
Hi @timebertt. Thank you very much for this very detailed review! I tried to address all comments and suggestions as best as possible. I kept the one about creating the `SecurityGroup` open but I think this one should be resolved too.
Also the integration tests are added as checks in https://github.com/stackitcloud/ske-ci-infra/pull/4666

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,98 @@
...
+	// Decode providerConfig (when present) and apply API defaults.
+	providerConfig := &stackitv1alpha1.SelfHostedShootExposureConfig{}
+	if exposure.Spec.ProviderConfig != nil {
+		if _, _, err := a.Decoder.Decode(exposure.Spec.ProviderConfig.Raw, nil, providerConfig); err != nil {
+			return nil, fmt.Errorf("error decoding providerConfig: %w", err)
+		}
+	}
+	a.Client.Scheme().Default(providerConfig)
+
+	if errs := validation.ValidateSelfHostedShootExposureConfig(providerConfig, field.NewPath("spec.providerConfig")); len(errs) > 0 {
```
timebertt: Nit: `field.NewPath` expects field path elements as separate params, i.e., `field.NewPath("spec", "providerConfig")`

pkg/stackit/client/loadbalancing.go
```diff
@@ -11,9 +11,14 @@ import (
...
 type LoadBalancingClient interface {
+	ProjectID() string
+
+	CreateLoadBalancer(ctx context.Context, payload loadbalancer.CreateLoadBalancerPayload) (*loadbalancer.LoadBalancer, error)
+	GetLoadBalancer(ctx context.Context, name string) (*loadbalancer.LoadBalancer, error)
 	ListLoadBalancers(ctx context.Context) ([]loadbalancer.LoadBalancer, error)
-	DeleteLoadBalancer(ctx context.Context, lbName string) error
-	GetLoadBalancer(ctx context.Context, id string) (*loadbalancer.LoadBalancer, error)
+	UpdateLoadBalancer(ctx context.Context, name string, payload loadbalancer.UpdateLoadBalancerPayload) (*loadbalancer.LoadBalancer, error)
+	UpdateLoadBalancerTargetPool(ctx context.Context, name, targetPool string, payload loadbalancer.UpdateTargetPoolPayload) (*loadbalancer.TargetPool, error)
```
timebertt: `UpdateLoadBalancerTargetPool` is no longer used

timebertt (commented):
This looks great! Thanks for addressing my suggestions :) 
Nits remaining from my side. I would ask the infra experts for a review now:

test/integration/selfhostedshootexposure/selfhostedshootexposure_suite_test.go
timebertt, line 1: Just noticed: you can probably omit the `stackit` directory in this file path.

timebertt (commented):
Generally
/lgtm

timebertt:
Rebased to resolve conflicts

test/integration/selfhostedshootexposure/selfhostedshootexposure_suite_test.go
timebertt: Done

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,98 @@
...
+	opts.NetworkID = infraStatus.Networks.ID
+
+	// Decode providerConfig (when present) and apply API defaults.
+	providerConfig := &stackitv1alpha1.SelfHostedShootExposureConfig{}
+	if exposure.Spec.ProviderConfig != nil {
+		if _, _, err := a.Decoder.Decode(exposure.Spec.ProviderConfig.Raw, nil, providerConfig); err != nil {
+			return nil, fmt.Errorf("error decoding providerConfig: %w", err)
+		}
+	}
+	a.Client.Scheme().Default(providerConfig)
```
breuerfelix:
```suggestion
	} else {
	    a.Client.Scheme().Default(providerConfig)
	}
```

pkg/controller/selfhostedshootexposure/options.go
```diff
@@ -0,0 +1,98 @@
...
+	// AllowedSourceRanges restricts which source CIDRs may reach the load balancer.
+	// Empty means unrestricted.
+	AllowedSourceRanges []string
+}
+
+func (a *Actuator) DetermineOptions(ctx context.Context, exposure *extensionsv1alpha1.SelfHostedShootExposure, cluster *extensionscontroller.Cluster, projectID string) (*Options, error) {
+	opts := &Options{
+		SelfHostedShootExposure: exposure,
+		ProjectID:               projectID,
+		ResourceName:            fmt.Sprintf("%s-exposure-%s", cluster.Shoot.Status.TechnicalID, exposure.Name),
```
breuerfelix: could potentially exceed the name limit which is 63 characters. Either strip the name and make sure there is no slash in the end or come up with a different naming scheme.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,329 @@
...
+	// an error so controller-runtime retries with backoff rather than 400ing the API.
+	if r.LoadBalancer.ExternalAddress == nil {
+		return fmt.Errorf("waiting for load balancer external address before updating")
+	}
+
+	// LB endpoint is PUT-only and requires sending the whole resource.
+	updated, err := r.LBClient.UpdateLoadBalancer(ctx, r.ResourceName, loadbalancer.UpdateLoadBalancerPayload{
+		Name:            &r.ResourceName,
+		Version:         r.LoadBalancer.Version,
+		ExternalAddress: r.LoadBalancer.ExternalAddress,
```
dergeberl, line 88: Just to mention it, because we discussed this morning: This is correct, sorry for the confusion!

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
```diff
@@ -0,0 +1,329 @@
...
+			AllowedSourceRanges: r.AllowedSourceRanges,
+		}
+	}
+	return opts
+}
+
+// loadBalancerNeedsUpdate reports whether any controller-managed field of the LB (targets,
+// plan, ACL) differs from the desired state. Caller must guarantee r.LoadBalancer is non-nil.
+func (r *Resources) loadBalancerNeedsUpdate(specTargets []loadbalancer.Target) bool {
+	return r.targetsNeedUpdate(specTargets) || r.planNeedsUpdate() || r.accessControlNeedsUpdate()
```
breuerfelix: It seems like the port of the SelfHostedExposureClass is not immutable, so we also need to check if it changed and reconcile the LoadBalancer accordingly.

jamand:
Oh no, there were changes to the Network Load Balancer API I noticed running the integration tests, the creation endpoint now just seems to silently strip away every label that has `stackit` in the name, that is both
``` 
    "cluster.stackit.cloud": "test",
    "exposure.stackit.cloud": "apiserver"
 ```
now just disappear. Labels as such can still be set, e.g. I tried with `"simple": "me"`.
The API docs state `Keys starting with 'stackit-' are system-reserved; users MUST NOT manage them.` which is by what I observed not completely true (they are much stricter).

Update: Seems to be a bug.

pkg/controller/selfhostedshootexposure/resources_loadbalancer.go
jamand: Good catch!

pkg/controller/selfhostedshootexposure/options.go
jamand: `exposure` -> `exp` + Added shortening for `exposure.Name` if it's too long, in that case shorten it so it + its 8-char hash (for uniqueness) fit into the 63 char limit.

jamand: Tested this: `test-exposure2-this-is-a-very-long-name-and-should-shorten` created a LB called `shoot--garden--s-eu01-000-exp-test-exposure2-this-is-a-a859ef9d`.

timebertt (approved):
Very nice, looks good from my side :)

timebertt:
pinging @stackitcloud/ske-infrastructure for approval

breuerfelix:
image
Thanks for implementing our suggestions!
Tested normal openstack + stackit shoot creation in an ondemand. Still working like a charm.

/approve

jamand:
Cool, thank you for testing! :)
I also re-ran the integration tests locally, they run through again, as LB fixed the API again. There was a bug in the GET/Listing of labels, so they were not returned (although they were properly set).
/unhold
I'll also merge https://github.com/stackitcloud/ske-ci-infra/pull/4666 which runs the integration tests for this one automatically on every new PR (the PR requires the tests/make target added here to exist).

PR #147, opened 2026-05-18
Add cloudprovider hash as annotation to STACKIT CSI

nschad:

/kind enhancement

Should add the hash of the `cloudprovider` secret to the CSI Deployment in the control-plane. Changes to the secret then will automatically recreate the deployment so the key in the application can be updated.

**Breaking changes**:

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -477,6 +477,7 @@ func (vp *valuesProvider) GetControlPlaneChartValues(
 	checksums[openstack.CloudProviderCSIDiskConfigName] = gardenerutils.ComputeChecksum(cpDiskConfigSecret.Data)
 	credentials, _ := vp.getCredentials(ctx, cp) // ignore missing credentials
 
+	// TODO: The comment says ignore, but is this actually ignored?! If the secret can't be retrieved??? Same as above.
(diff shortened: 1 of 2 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -477,6 +477,7 @@ func (vp *valuesProvider) GetControlPlaneChartValues(
 	checksums[openstack.CloudProviderCSIDiskConfigName] = gardenerutils.ComputeChecksum(cpDiskConfigSecret.Data)
 	credentials, _ := vp.getCredentials(ctx, cp) // ignore missing credentials
 
+	// TODO: The comment says ignore, but is this actually ignored?! If the secret can't be retrieved??? Same as above.
```
Kumm-Kai: lol :D If it's not ignored by the function itself then It's wrong I guess. Generally it doesn't sound good to ignore missing credentials

nschad: I couldn't find any ignore. I really don't understand that comment @Kumm-Kai

nschad: Determinted probably this is just a copy-pasta fail from above.

Kumm-Kai (approved):

PR #148, opened 2026-05-20
refactor registryCache mirror code into exported type

hown3d:

/kind enhancement

Allows to reuse the registry-cache implementation from outside this module. 

/cc @dergeberl 

**Breaking changes**:


pkg/webhook/controlplane/ensurer.go
```diff
@@ -36,15 +36,17 @@ import (
 // NewEnsurer creates a new controlplane ensurer.
 func NewEnsurer(regCaches []config.RegistryCacheConfiguration, logger logr.Logger) genericmutator.Ensurer {
```
dergeberl, line 38: Cant you just use the `NewEnsurer` to use it outside of this pkg?

hown3d: No, since we want to use exactly only the registry cache logic. If the provider decides to add more functionality into the `EnsureAdditionalProvisionFiles` we would also include such logic.

dergeberl (approved):

PR #150, opened 2026-05-21
Use CA in cloudprofile for STACKIT clients

nschad:

/kind enhancement

Removes the custom-ca secret and instead uses the CA from the `Cloudprofile` for all stackit components. OpenStack is unaffected.

Fixes #

**Breaking changes**:


charts/gardener-extension-provider-stackit/templates/deployment.yaml
```diff
@@ -144,10 +144,6 @@ spec:
         configMap:
           name: {{ include "name" . }}-configmap
           defaultMode: 420
-      - name: custom-certs
```
nschad, line 147: I assume if I remove this the openstack client breaks

nschad: Nevermind the CA is directly passed in the openstack client. Should ™️ work

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/deployment-csi-driver-controller.yaml
```diff
@@ -341,15 +337,11 @@ spec:
...
           name: {{ .Values.prefix }}-cloud-provider-config
-      # Host certificates are mounted to accommodate OpenStack endpoints that might be served with a certificate
-      # signed by a CA that is not globally trusted.
-      - name: etc-ssl
-        hostPath:
-          path: /etc/ssl
+      - name: stackit-ca
+        secret:
+          optional: true
+          secretName: stackit-ca-bundle
```
nschad: We could wrap this with ```{{if```. If somebody currently (for wahtever reason) enables the caBundle then the pods needs to be restarted manually.

charts/internal/cloud-provider-config/values.yaml
```diff
@@ -1,4 +1,5 @@
 stackitOnly: false
+cloudProfileCABundleB64: ""
```
breuerfelix:
```suggestion
CABundleBase64: ""
```

pkg/webhook/controlplane/ensurer.go
```diff
@@ -83,6 +88,23 @@ func (e *ensurer) EnsureMachineControllerManagerDeployment(ctx context.Context,
 
 		apiEndpoints := ptr.Deref(cloudProfileConfig.APIEndpoints, stackitv1alpha1.APIEndpoints{})
 
+		if cloudProfileConfig.CABundle != nil {
+			machineDeployment.Spec.Template.Spec.Volumes = append(machineDeployment.Spec.Template.Spec.Volumes, corev1.Volume{
+				Name: "stackit-ca",
```
breuerfelix: use a variable for the name

breuerfelix (changes requested):
Add tests for as much as possible. E.g. webhook, factory and maybe valuesprovider

charts/internal/cloud-provider-config/values.yaml
nschad: Fine

pkg/apis/stackit/v1alpha1/types_cloudprofile.go
```diff
@@ -31,10 +31,6 @@ type CloudProfileConfig struct {
 	// APIEndpoints contains API endpoints for various services (e.g., "LoadBalancer", "IaaS").
 	// +optional
 	APIEndpoints *APIEndpoints `json:"apiEndpoints,omitempty"`
-	// CABundle is the CA certificate bundle for API endpoints.
-	// This field is currently ignored and reserved for future use.
-	// +optional
-	CABundle *string `json:"caBundle,omitempty"`
```
breuerfelix, line 37: I don't know if we can just remove this field. We need to deprecate it first.
Since `CloudProfile` decoding is strict, this will fail for shoots that still have the CA set.
I remember that i wanted to add the DNSEndpoint to the cloudProfile but then the yawol extension was crashlooping since it didn't expect this field to be present.
That means that if the `CABundle` is still present, this controller would fail decoding it. 
I think we should:
1. deprecate that field
2. make sure no cloudprofile has the cabundle set
3. remove the cabundle field

pkg/controller/controlplane/valuesprovider.go
breuerfelix, line 949: We also need to add a checksum of the new stackit-ca here, we also need to make sure that the other pods that use the secret has the corresponding checksum in order for ca rotation to work properly

breuerfelix (changes requested):

pkg/controller/controlplane/valuesprovider.go
nschad: CA rotation? That's not really a thing to be honest, atleast in this case

pkg/apis/stackit/v1alpha1/types_cloudprofile.go
nschad: Or we could just mark this as breaking and then inform everybody that the `CABundle` needs to be removed. That has happen anyway, why not now?

breuerfelix: But this has to be two step anyway.
We cannot remove the caBundle until this branch is rolled out, otherwise the caBundle would be broken.
But we cannot rollout these Changes until the caBundle is removed.
You would have to rollout both in parallel and i don't think that this is possible due to async reconciliation.

nschad: restored

dergeberl: Agree that deprecate is the cleanest way.

> otherwise the caBundle would be broken
Not 100% true as the field never did anything, it is most likely not set anywhere.

pkg/stackit/client/factory.go
```diff
@@ -72,10 +83,28 @@ func (f factory) DNS(ctx context.Context, c client.Client, secretRef corev1.Secr
...
-	return NewDNSClient(ctx, f.StackitAPIEndpoints, credentials)
+	return NewDNSClient(ctx, f.StackitAPIEndpoints, credentials, f.CABundleB64)
 }
 
-func clientOptions(endpoints stackitv1alpha1.APIEndpoints, credentials *stackit.Credentials) []sdkconfig.ConfigurationOption {
+// newHTTPClientWithCustomCA creates an http.Client with a custom CA
+func newHTTPClientWithCustomCA(caBundle []byte) (*http.Client, error) {
+	caCertPool, err := x509.SystemCertPool()
+	if err != nil {
+		// we could also fall back here and use an empty pool via x509.NewCertPool()
```
dergeberl, line 93: I think erroring is fine but we should customize the error to state what is broken and where does it come from.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -135,6 +135,7 @@ var (
 		Objects: []*chart.Object{
 			{Type: &corev1.Secret{}, Name: openstack.CloudProviderConfigName},
 			{Type: &corev1.Secret{}, Name: openstack.CloudProviderDiskConfigName},
+			{Type: &corev1.Secret{}, Name: "stackit-ca-bundle"},
```
dergeberl: lets also use a const here as we also do this for most of the rest.

charts/internal/cloud-provider-config/templates/stackit-ca-bundle.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.CABundle }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: stackit-ca-bundle
```
dergeberl: As the `CloudProfile.Spec.CABundle` field is no STACKIT exclusive I am not sure if we should name it differently (like `cloudprofile-ca-bundle`), WDYT?

nschad: sure!

nschad: I've renamed this. Do you also want me to rename the "internal-path". For example it's called `stackit-ca.crt` everywhere. I thought its okay since this is the "stackit" extension. So 🤷

pkg/apis/stackit/v1alpha1/types_cloudprofile.go
nschad: Yup Done.

pkg/controller/controlplane/valuesprovider.go
breuerfelix: True, but it also doesn't hurt right? If, for whatever reason, the ca gets rotated, it would be fully automated then :) I am also fine with not adding the checksum but since its just one line, why not :D

charts/internal/cloud-provider-config/templates/stackit-ca-bundle.yaml
breuerfelix: I think that is a good idea to also rename the key of the cert.

nschad: Done.

breuerfelix (approved):

PR #161, opened 2026-05-27
Remove fork from mcm provider openstack and use newest release

viragvoros:
/kind enhancement

We want to get rid of the fork for machine-controller-manager-provider-openstack, therefore the necessary changes were contributed upstream. See https://github.com/gardener/machine-controller-manager-provider-openstack/releases/tag/v0.27.0

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -90,7 +90,6 @@
       matchPackageNames: [
         'europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack',
       ],
-      ignoreUnstable: false,
     },
     {
       groupName: 'Internal',
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -86,12 +86,6 @@
       ],
     },
     {
-      groupName: 'machine-controller-manager-provider-openstack',
-      matchPackageNames: [
-        'europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack',
-      ],
-    },
-    {
       groupName: 'Internal',
       matchFileNames: [
         'charts/internal/**/requirements.yaml',
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -88,9 +88,8 @@
     {
       groupName: 'machine-controller-manager-provider-openstack',
       matchPackageNames: [
-        'registry.ske.stackit.cloud/stackitcloud/machine-controller-manager-provider-openstack',
+        'europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack',
       ],
-      // Enable updating fork releases/tags, e.g., v0.24.1-ske-1
       ignoreUnstable: false,
     },
     {
--- a/imagevector/images.yaml
+++ b/imagevector/images.yaml
@@ -44,8 +44,8 @@ images:
 
 - name: machine-controller-manager-provider-openstack
   sourceRepository: github.com/gardener/machine-controller-manager-provider-openstack
-  repository: registry.ske.stackit.cloud/stackitcloud/machine-controller-manager-provider-openstack
-  tag: "v0.25.0-ske-1"
+  repository: europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack
+  tag: "v0.27.0"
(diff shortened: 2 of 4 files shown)


imagevector/images.yaml
```diff
@@ -44,8 +44,8 @@ images:
 
 - name: machine-controller-manager-provider-openstack
   sourceRepository: github.com/gardener/machine-controller-manager-provider-openstack
-  repository: registry.ske.stackit.cloud/stackitcloud/machine-controller-manager-provider-openstack
-  tag: "v0.25.0-ske-1"
+  repository: europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack
```
maboehm, line 47: do we also need to customize the flag `--resource-exhausted-retry` that was introduced in that change as well?

.github/renovate.json5
```diff
@@ -88,10 +88,8 @@
     {
       groupName: 'machine-controller-manager-provider-openstack',
       matchPackageNames: [
-        'registry.ske.stackit.cloud/stackitcloud/machine-controller-manager-provider-openstack',
+        'europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack',
       ],
-      // Enable updating fork releases/tags, e.g., v0.24.1-ske-1
-      ignoreUnstable: false,
     },
```
maboehm: I think that entire group can now be removed.

imagevector/images.yaml
viragvoros: It is an optional flag, for now I added it (I am a bit unsure about the implementation), but we would need to discuss what would be an appropriate retry period for our case.

dergeberl: I think that is the reason why we contibuted this feature to the machine-controller-manager. It was set in our fork of machine-controller-manager to 30min https://github.com/stackitcloud/machine-controller-manager/commit/a8329ddac4673d049da5f2f60659f13ba7e8a3bd

dergeberl: I think we also should set this for the STACKIT MCM because there we also want to avoid recreate everything in case of resource exhausted or?
cc @Kumm-Kai you did the changes in the machine-controller-manager fork. WDYT?

Kumm-Kai: I'm not sure about the implementation of the STACKIT MCM, do we still create the volume **before** the server or do we do it in the server create call?
If it's the same as in OpenStack MCM then the flag would be useful otherwise we don't need it, because the STACKIT API wouldn't create the volume when the server cannot be scheduled.

breuerfelix: In the STACKIT MCM we let IaaS create the volume for us. It is just one call that does it all.

dergeberl: @Kumm-Kai But it also dont help to recreate the node more frequently if the resource is exhausted or?

pkg/webhook/controlplane/ensurer.go
```diff
@@ -98,6 +98,13 @@ func (e *ensurer) EnsureMachineControllerManagerDeployment(ctx context.Context,
 		}
 	}
 
+	if !feature.UseStackitMachineControllerManager(cluster) {
+		sidecarContainer.Args = extensionswebhook.EnsureStringWithPrefix(
+			sidecarContainer.Args,
+			"--resource-exhausted-retry=", "30m",
+		)
+	}
```
dergeberl:
```suggestion
    // avoid unnecessary reconciles if a recourse exhausted. Specially needed for machine-controller-manager-provider-openstack
    // to avoid the creation of many volumes. But also on machine-controller-manager-provider-stackit this limits the amount of requests.
	sidecarContainer.Args = extensionswebhook.EnsureStringWithPrefix(
		sidecarContainer.Args,
		"--resource-exhausted-retry=", "30m",
	)
```

I quickly discussed with @Kumm-Kai that it still makes sense to increase the time for reconciles in that case for provider-stackit.

imagevector/images.yaml
dergeberl: see: https://github.com/stackitcloud/gardener-extension-provider-stackit/pull/161/changes#r3348345642

viragvoros:
Hi @breuerfelix, could you take a final quick look? :)

breuerfelix (approved):

PR #162, opened 2026-06-01
fix: Use BuildResourceName to build a valid DNS1123 compliant bastion name

nschad:

/kind bug

The generated bastionName can be bigger than allowed (max 63 characters). Therefore we introduce the same mechanism as in the exposure classes for the generation of the bastionName.

This PR makes the `BuildResourceName` a bit more generic and moves the helper function to the stackit client package.

Fixes #

**Breaking changes**:

### DIFF
--- a/pkg/controller/bastion/options.go
+++ b/pkg/controller/bastion/options.go
@@ -10,6 +10,7 @@ import (
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	stackitclient "github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/stackit/client"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/apis/stackit/helper"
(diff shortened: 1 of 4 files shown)


pkg/controller/bastion/options.go
```diff
@@ -43,10 +44,15 @@ type Options struct {
 }
 
 func (a *Actuator) DetermineOptions(ctx context.Context, bastion *extensionsv1alpha1.Bastion, cluster *extensionscontroller.Cluster, projectID string) (*Options, error) {
+	bastionName := fmt.Sprintf("%s-bastion-%s", cluster.Shoot.Status.TechnicalID, bastion.Name)
+	if len(bastionName) > 63 {
```
dergeberl: This check is done also in the `BuildResourceName` or? `utilvalidation.DNS1123LabelMaxLength` is also `63`.

pkg/stackit/client/utils.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"strings"
+
+	gardenerutils "github.com/gardener/gardener/pkg/utils"
+	utilvalidation "k8s.io/apimachinery/pkg/util/validation"
+)
+
+const (
+	// resourceNameHashLength is the number of hex chars from SHA-256(exposure.Name) appended when
```
dergeberl:
```suggestion
	// resourceNameHashLength is the number of hex chars from SHA-256 appended when
```

pkg/stackit/client/utils.go
```diff
@@ -0,0 +1,31 @@
...
+	// the unabridged name would exceed validation.DNS1123LabelMaxLength. 8 hex chars = 32 bits,
+	// plenty of collision resistance for "names sharing a prefix on the same shoot".
+	resourceNameHashLength = 8
+)
+
+// BuildResourceName returns "<technicalID>-<resourceNameInfix>-<resourceName>" if it fits within
+// apivalidation.DNS1123LabelMaxLength, otherwise truncates the exposure name and appends an
```
dergeberl:
```suggestion
// apivalidation.DNS1123LabelMaxLength, otherwise truncates the resource name and appends an
```

pkg/controller/bastion/options.go
hown3d: I agree, No need to have this check here.
Just always use BuildResourceName

hown3d (approved):

dergeberl (approved):

PR #175, opened 2026-06-12
CSI compatibility mode

ftl:

/kind enhancement

This change allows to deploy two separate instances of the STACKIT CSI driver, on in "normal" mode, and the second one in "compatibility" mode. This is part of the migration from cinder to the STACKIT CSI driver. 

Fixes #

This change requires the enhancement https://github.com/stackitcloud/cloud-provider-stackit/pull/161 to be available in the cloud-provider-stackit CSI driver plugin.

**Breaking changes**:


charts/internal/shoot-system-components/charts/stackit-blockstorage-csi-driver/templates/daemonset.yaml
```diff
@@ -48,6 +48,12 @@ spec:
         {{- end }}
         - --v=2
         - --provide-controller-service=false
+        {{- if .Values.csi.enableCompatibilityMode }}
+        - --legacy-storage-mode=true
+        {{- end }}
+        {{- if .Values.csi.blockLegacyCreation }}
+        - --legacy-volume-creation=false
+        {{- end }}
```
nschad:
```suggestion
```

This needs to be in seed-controlplane chart

charts/internal/shoot-system-components/charts/stackit-blockstorage-csi-driver/values.yaml
```diff
@@ -3,6 +3,12 @@ driverName: block-storage.csi.stackit.cloud
 
 rescanBlockStorageOnResize: "true"
 
+healthzPort: 9908
+
+csi:
+  enableCompatibilityMode: false
+  blockLegacyCreation: false
```
nschad:
```suggestion
```

See above.

pkg/controller/controlplane/csi_compatibility.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/charts"
+	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/imagevector"
+	stackitv1alpha1 "github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/apis/stackit/v1alpha1"
+	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/openstack"
+)
+
+const (
+	csiDriverChartName      = "stackit-blockstorage-csi-driver"
+	csiCompatibilityPrefix  = "stackit-csi-compat"
```
nschad: maybe find a better name for this?

pkg/controller/controlplane/csi_compatibility.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	chartValues["images"] = imageMap
+
+	chartValues["healthzPort"] = 9909
+	csiValues := map[string]any{
+		"enableCompatibilityMode": true,
+	}
+	if blockLegacyCreation {
+		csiValues["blockLegacyCreation"] = true
+	}
```
nschad:
```suggestion
```

Needs to be set on the seed :)

nschad (dismissed):

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/csi-snapshot-controller-poddisruptionbudget.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.csi.enableCompatibilityMode }}
```
dergeberl, line 1: Why do we dont deploy the `PodDisruptionBudget` when enable the CompatibilityMode

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/csi-snapshot-controller-vpa.yaml
```diff
@@ -1,4 +1,4 @@
----
+{{- if not .Values.csi.enableCompatibilityMode }}
```
dergeberl, line 1: Why do we dont deploy the `VerticalPodAutoscaler` when enable the CompatibilityMode

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/deployment-csi-snapshot-controller.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.csi.enableCompatibilityMode }}
```
dergeberl, line 1: This is not needed as the `snapshot-controller` dont use the driverName right? Snapshots can be still created for both drivernames?

charts/internal/shoot-system-components/charts/stackit-blockstorage-csi-driver/templates/daemonset.yaml
```diff
@@ -17,7 +17,11 @@ spec:
     metadata:
       annotations:
         checksum/configmap-cloud-provider-config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
-        node.gardener.cloud/wait-for-csi-node-stackit: block-storage.csi.stackit.cloud
+        {{- if eq .Values.driverName "cinder.csi.openstack.org" }}
```
dergeberl: We could use the `enableCompatibilityMode` instead if the driver here
```suggestion
        {{- if .Values.csi.enableCompatibilityMode }}
```

dergeberl: It is then also needed to switch the annotations in the if/else

pkg/apis/stackit/v1alpha1/types_controlplane.go
```diff
@@ -58,7 +58,8 @@ type Storage struct {
 }
 
 type CSI struct {
-	Name string `json:"name"`
+	Name              string `json:"name"`
+	CompatibilityMode string `json:"compatibilityMode,omitempty"`
```
dergeberl: Can we add a description/comment here what this fields are and what the default is?

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/csi-snapshot-controller-poddisruptionbudget.yaml
nschad: @dergeberl We only need one snapshot controller per cluster. It would just be duplicated doing nothing.

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/csi-snapshot-controller-vpa.yaml
nschad: Same as above. One snapshot-controller is suffice.

charts/internal/seed-controlplane/charts/stackit-blockstorage-csi-driver/templates/deployment-csi-snapshot-controller.yaml
nschad: Yes, correct. The leader of the controller is also hard-coded so it wouldn't even bring you anything if you have multiple controllers running.

charts/internal/shoot-system-components/charts/stackit-blockstorage-csi-driver/templates/daemonset.yaml
```diff
@@ -17,7 +17,11 @@ spec:
     metadata:
       annotations:
         checksum/configmap-cloud-provider-config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
-        node.gardener.cloud/wait-for-csi-node-stackit: block-storage.csi.stackit.cloud
+        {{- if .Values.csi.enableCompatibilityMode }}
```
hasit97, line 20: why is the `if else` block here if there is only `node.gardener.cloud/wait-for-csi-node-stackit` ?

pkg/controller/controlplane/csi_compatibility.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+type CompatCSICompatibilityHandler struct {
+	client   client.Client
+	renderer chartrenderer.Interface
+}
+
+func (ch *CompatCSICompatibilityHandler) HandleSeedCSICompatibility(ctx context.Context, namespace string, version string, cpConfig *stackitv1alpha1.ControlPlaneConfig, controlPlaneValues map[string]any) error {
+	compatibilityMode := getCSICompatibilityMode(cpConfig)
+	switch compatibilityMode {
```
hasit97: if compat mode != compat or compat block -> delete + early return ?

charts/internal/shoot-system-components/charts/stackit-blockstorage-csi-driver/templates/daemonset.yaml
ftl: As documented in https://github.com/gardener/gardener/blob/master/docs/usage/advanced/node-readiness.md#marking-node-critical-components the annotation name must include either "openstack" or "stackit". No way to get rid of the if/else block.

breuerfelix:
Maybe we should also validate the new compatiblity field in `pkg/apis/stackit/validation/controlplane.go` ? Since currently, if we would introduce a typo like `compatblok`, we would remove the whole compatibility layer without noticing since we fallback to no compat at all.

pkg/controller/controlplane/csi_compatibility.go
```diff
@@ -0,0 +1,190 @@
...
+	return client.IgnoreNotFound(managedresources.DeleteForShoot(ctx, ch.client, namespace, csiCompatShootChartName))
+}
+
+// composeCompatibilityChartValues returns a copy of the given values map merged with the csiStackitValues on topLevel.
+// Basically removes the openstack.CSISTACKITControllerName key
+func composeCompatibilityChartValues(values map[string]any) map[string]any {
+	if values == nil {
+		return map[string]any{}
+	}
+	csiStackitValues, ok := values[openstack.CSISTACKITControllerName].(map[string]any)
```
breuerfelix: This is true for the Seed Chart but not for the Shoot Chart.
When calling this function in the Shoot Chart, we need to use the key `openstack.CSISTACKITNodeName`.
Maybe it makes sense to get rid of the function in that case and just inline it directly in both cases with the respective key.

nschad: It's the same for both. There is `CSISTACKITControllerName` and `CSISTACKITNodeName` but the both have the same value "stackit-blockstorage-csi-driver" which is also the name of the chart

nschad (approved):

PR #182, opened 2026-06-19
feat: make root disk of bastion configurable

nschad:
/kind enhancement

Our "coreos" image size was recently increased to 13GiB (see [upstream PR](https://github.com/flatcar/scripts/pull/3027) or https://github.com/flatcar/Flatcar/issues/1917). In order to make the bastion controller a bit more flexible I've leverage the provider-specific `CloudprofileConfig` to allow for custom `RootDiskSizes`.

Additionally we moved the status check immediately after `reconcileServer` as the VM creation is async and we should first check if the server was created properly before doing further operations on said server.

Fixes #

**Breaking changes**:


pkg/controller/bastion/options.go
```diff
@@ -37,12 +37,19 @@ type Options struct {
...
+	RootDiskSize int64
 	// Machine type and image for the Bastion, determined from CloudProfile (spec.bastion and spec.providerConfig.machineImages).
 	MachineType, ImageID string
 	// Network and security group used by shoot workers, determined from Infrastructure.status.providerStatus.
 	NetworkID, WorkerSecurityGroupID string
 }
 
+const (
+	// RootDiskSize is the default root disk size for the bastion server
+	RootDiskSize = int64(25)
```
hown3d: nit: rename to DefaultRootDiskSize

hown3d (commented):
Thanks for this change!

just a small nit :)

pkg/controller/bastion/options.go
nschad: Done. Thanks

robinschneider (approved):
/lgtm

nschad:

PR #202, opened 2026-07-07
Use fake client instead of mock client for valuesprovider_test.go

breuerfelix:

/kind test

We want to migrate away from the mock client so this PR rewrites the test to leverage fake clients.

**Breaking changes**:


pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
+	rescanBlockStorageOnResize := true
+	ignoreVolumeAZ := true
+	nodeVolumeAttachLimit := int32(25)
+
+	return &stackitv1alpha1.CloudProfileConfig{
+		KeyStoneURL:                authURL,
+		RequestTimeout:             testRequestTimeout,
+		RescanBlockStorageOnResize: &rescanBlockStorageOnResize,
+		IgnoreVolumeAZ:             &ignoreVolumeAZ,
+		NodeVolumeAttachLimit:      &nodeVolumeAttachLimit,
```
dergeberl: can be done inline 
```suggestion
		RescanBlockStorageOnResize: new(true),
		IgnoreVolumeAZ:             new(true),
		NodeVolumeAttachLimit:      new(int32(25)),
```

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
+	}
+}
 
-		cidr                             = "10.250.0.0/19"
-		rescanBlockStorageOnResize       = true
-		ignoreVolumeAZ                   = true
-		nodeVoluemAttachLimit      int32 = 25
+func baseCluster() *extensionscontroller.Cluster {
+	podCIDR := "10.250.0.0/19"
```
dergeberl: can also be done inline.

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
+	cluster := baseCluster()
+	networkType := "calico"
+	cluster.Shoot.Spec.Networking = &gardencorev1beta1.Networking{
+		Type: &networkType,
+		ProviderConfig: &runtime.RawExtension{
+			Raw: []byte(`{"overlay":{"enabled":false}}`),
+		},
+		Pods: cluster.Shoot.Spec.Networking.Pods,
+	}
+	cluster.Shoot.Spec.Kubernetes.Version = "1.31.1"
```
dergeberl: Is the version relevant in that case?

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
+				Networking: &gardencorev1beta1.Networking{
+					Pods: &podCIDR,
 				},
-			},
-			CloudProfile: &gardencorev1beta1.CloudProfile{
-				Spec: gardencorev1beta1.CloudProfileSpec{
-					ProviderConfig: &runtime.RawExtension{
-						Raw: cloudProfileConfigJSON,
+				Kubernetes: gardencorev1beta1.Kubernetes{
+					Version: "1.29.0",
```
dergeberl: does it make sense to load this dynamically?

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
-			"applicationCredentialID":     "",
-			"applicationCredentialSecret": "",
-			"applicationCredentialName":   "",
-			"internalNetworkName":         technicalID,
-		}
-
 		BeforeEach(func() {
-			// Disable STACKIT feature flags for OpenStack-only tests
 			DeferCleanup(testutils.WithFeatureGate(feature.MutableGate, feature.UseSTACKITAPIInfrastructureController, false))
 			DeferCleanup(testutils.WithFeatureGate(feature.MutableGate, feature.UseSTACKITMachineControllerManager, false))
```
dergeberl, line 445: I know you just adopted the tests to the fake client, but we should also check the config chart values when STACKIT Only (this should look different.)

dergeberl: We can also add a TODO here and do this in a later PR, together with the other things you found.

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
-		stackitPodIdentityWebhookChartSeedValues := map[string]any{
-			"enabled":      false,
-			"replicaCount": 1,
-			"webhook": map[string]any{
-				"tlsSecretName": stackitPodIdentityWebhookServerName,
-			},
-		}
+			ccmValues := chartValues(values, openstack.CloudControllerManagerName)
+			Expect(ccmValues).To(BeComparableTo(map[string]any{
+				"enabled":           false,
```
dergeberl: It would be in my opinion enough to just check if enabled=false

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
-			Expect(fakeClient.Create(context.TODO(), &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "cloud-controller-manager-server", Namespace: namespace}})).To(Succeed())
-			Expect(fakeClient.Create(context.TODO(), &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: stackitPodIdentityWebhookServerName, Namespace: namespace}})).To(Succeed())
-
-			// This call is made for emergency Loadbalancer API access.
-			// It will return a NotFound error by default to not interfere with existing tests.
-			// Returning this error effectively disables the emergency access feature.
-			c.EXPECT().Get(ctx, types.NamespacedName{Name: LoadBalancerEmergencyAccessSecretName, Namespace: namespace}, &corev1.Secret{}).Return(
-				errors.NewNotFound(schema.GroupResource{Resource: "secret"}, LoadBalancerEmergencyAccessSecretName))
-		})
+			expectedSTACKITCCMConfig := expectedSTACKITCCMConfig("kubernetes.io", nil)
```
dergeberl: Can we do this inline down below also to avoid a shadow declaration?

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
-				"example.com",
-				"example.com/cluster",
-			),
+			Entry("default kubernetes.io domain", "kubernetes.io"),
+			Entry("custom ske.stackit.cloud domain", "ske.stackit.cloud"),
+			Entry("custom example.com domain", "example.com"),
 		)
 
-		DescribeTable("topologyAwareRoutingEnabled value",
+		DescribeTable("supports topology-aware-routing compatibility combinations",
```
dergeberl: TBH I dont get what this test is testing. What does this have to do with the STACKIT CSI?

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -128,1017 +127,938 @@ func controlPlane(floatingPoolID string, cfg *stackitv1alpha1.ControlPlaneConfig
...
+				"config": map[string]any{
+					"region":           "eu01",
+					"stackitProjectID": "foo",
+					"stackitNetworkID": "network-acbd1234",
+				},
+			}))
+		})
 
-			values, err := vp.GetControlPlaneChartValues(ctx, cp, cluster, fakeSecretsManager, checksums, false)
+		It("omits ALB controller values when the harness disables ALB deployment", func() {
```
dergeberl: harness?
omits ALB controller values when the config disables ALB deployment

breuerfelix: this is a leftover from the approach i used before, i constructed a "testharness" struct that was responsible to keep all objects and create them.

cleaned up

pkg/controller/controlplane/valuesprovider_test.go
breuerfelix: What variable are we shadowing here?
We need the expectedConfig as a variable since it is used for the checksum aswell.

pkg/controller/controlplane/valuesprovider_test.go
breuerfelix: Me neither. I just ported it but haven't thought about if it makes sense or not.
Should I just drop the test?

pkg/controller/controlplane/valuesprovider_test.go
breuerfelix: Do you mean setting the kubernetes version dynamically?

pkg/controller/controlplane/valuesprovider_test.go
breuerfelix: added the test case

dergeberl (approved):

PR #203, opened 2026-07-08
Add validation webhook for shoot object in logging mode

dergeberl:

/kind enhancement

We already have validation functions for the `ControlPlaneConfig` and `InfrastructureConfig` but we had no validation webhook for the shoot object. This PR adds the webhook.

It also fixes the `InfrastructureConfig` as this was not working, as it is possible to habe no `InfrastructureConfig.Networks.Workers` if `InfrastructureConfig.Networks.ID` is set. 

I created a cluster with and without SNA via the ske-api.

We want to use this for the ALB integration #54 

Please also check the validation itself as this was not used before and needed already adoption in the `InfrastructureConfig` part.

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -70,7 +70,9 @@ func (s *shoot) Validate(_ context.Context, newObj, oldObj client.Object) error 
 	}
 
 	// only log errors for now to be able to check the logs after a while before actually fail validation.
-	logger.Error(allErrs.ToAggregate(), "Shoot would have failed validation", "name", shoot.Name, "namespace", shoot.Namespace)
+	if allErrs.ToAggregate() != nil {
+		logger.Error(allErrs.ToAggregate(), "Shoot would have failed validation", "name", shoot.Name, "namespace", shoot.Namespace)
+	}
 
 	return nil
 }
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -1,0 +1,68 @@
+// SPDX-FileCopyrightText: SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package validator
+
+import (
+	"context"
+	"fmt"
+
+	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
+	"github.com/gardener/gardener/pkg/apis/core"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/apis/stackit/helper"
+	stackitvalidation "github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/apis/stackit/validation"
+)
+
+// NewShootValidator returns a new instance of a shoot validator.
+func NewShootValidator(mgr manager.Manager) extensionswebhook.Validator {
+	return &shoot{}
+}
+
+type shoot struct{}
+
+// Validate validates the given shoot objects.
+func (s *shoot) Validate(_ context.Context, newObj, oldObj client.Object) error {
+	shoot, ok := newObj.(*core.Shoot)
+	if !ok {
+		return fmt.Errorf("wrong object type %T", newObj)
+	}
+
+	if shoot.Spec.Provider.ControlPlaneConfig == nil {
+		return nil
+	}
+
+	cpConfig, err := helper.ControlPlaneConfigFromRawExtension(shoot.Spec.Provider.ControlPlaneConfig)
+	if err != nil {
+		return err
+	}
+
+	infraConfig, err := helper.InfrastructureConfigFromRawExtension(shoot.Spec.Provider.InfrastructureConfig)
+	if err != nil {
+		return err
+	}
+
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfig(cpConfig, shoot.Spec.Kubernetes.Version, field.NewPath("spec").Child("providerConfig"))...)
+
+	allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfig(infraConfig, shoot.Spec.Networking.Nodes, field.NewPath("spec").Child("providerConfig").Child("infrastructureConfig"))...)
+
+	if oldObj != nil {
+		oldShoot, ok := oldObj.(*core.Shoot)
+		if !ok {
+			return fmt.Errorf("wrong object type %T for old object", oldObj)
+		}
+		oldInfraConfig, err := helper.InfrastructureConfigFromRawExtension(oldShoot.Spec.Provider.InfrastructureConfig)
+		if err != nil {
+			return err
+		}
+		allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfigUpdate(oldInfraConfig, infraConfig, field.NewPath("spec").Child("providerConfig").Child("infrastructureConfig"))...)
+	}
+
+	return allErrs.ToAggregate()
+}
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -69,5 +69,8 @@ func (s *shoot) Validate(_ context.Context, newObj, oldObj client.Object) error 
 		allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfigUpdate(oldCpConfig, cpConfig, field.NewPath("spec").Child("provider").Child("controlPlaneConfig"))...)
 	}
 
-	return allErrs.ToAggregate()
+	// only log errors for now to be able to check the logs after a while before actually fail validation.
+	logger.Error(allErrs.ToAggregate(), "Shoot would have failed validation", "name", shoot.Name, "namespace", shoot.Namespace)
+
+	return nil
 }
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -11,6 +11,7 @@ import (
 	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
 	"github.com/gardener/gardener/pkg/apis/core"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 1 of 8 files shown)


pkg/admission/validator/shoot.go
```diff
@@ -0,0 +1,68 @@
...
+	if oldObj != nil {
+		oldShoot, ok := oldObj.(*core.Shoot)
+		if !ok {
+			return fmt.Errorf("wrong object type %T for old object", oldObj)
+		}
+		oldInfraConfig, err := helper.InfrastructureConfigFromRawExtension(oldShoot.Spec.Provider.InfrastructureConfig)
+		if err != nil {
+			return err
+		}
+		allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfigUpdate(oldInfraConfig, infraConfig, field.NewPath("spec").Child("providerConfig").Child("infrastructureConfig"))...)
```
nschad: why is this on the oldObj?

pkg/apis/stackit/validation/controlplane.go
```diff
@@ -21,13 +21,10 @@ var (
...
-func ValidateControlPlaneConfig(controlPlaneConfig *stackitv1alpha1.ControlPlaneConfig, infraConfig *stackitv1alpha1.InfrastructureConfig, version string, fldPath *field.Path) field.ErrorList {
-	allErrs := field.ErrorList{}
+func ValidateControlPlaneConfig(controlPlaneConfig *stackitv1alpha1.ControlPlaneConfig, version string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{} // nolint:prealloc // size is not known yet
 
-	if controlPlaneConfig.CloudControllerManager != nil {
-		allErrs = append(allErrs, featurevalidation.ValidateFeatureGates(controlPlaneConfig.CloudControllerManager.FeatureGates, version, fldPath.Child("cloudControllerManager", "featureGates"))...)
-		allErrs = append(allErrs, validateCloudController(controlPlaneConfig.CloudControllerManager, fldPath.Child("cloudControllerManager"))...)
-	}
+	allErrs = append(allErrs, validateCloudController(controlPlaneConfig.CloudControllerManager, version, fldPath.Child("cloudControllerManager"))...)
```
nschad, line 27: `if controlPlaneConfig.CloudControllerManager`

ah okay CloudControllerManager is defaulted

pkg/apis/stackit/validation/infrastructure.go
```diff
@@ -43,23 +34,38 @@ func ValidateInfrastructureConfig(infra *stackitv1alpha1.InfrastructureConfig, n
 		allErrs = append(allErrs, cidrvalidation.ValidateCIDRParse(workerCIDR)...)
 		//nolint:staticcheck // SA1019: needed for migration purposes
 		allErrs = append(allErrs, cidrvalidation.ValidateCIDRIsCanonical(networksPath.Child("worker"), infra.Networks.Worker)...)
+		if infra.Networks.ID != nil {
```
nschad, line 37: where was the different again between `worker` and `workers`. Wasn't one deprecated?

pkg/admission/validator/shoot.go
dergeberl: It has both it validates updates and checks for immutable fields.

pkg/apis/stackit/validation/infrastructure.go
dergeberl: Yes `worker` is depricated.

pkg/apis/stackit/validation/controlplane.go
dergeberl: The nil check is still done. It was also done in `validateCloudController` that is why is removed it here

dergeberl:
/hold
needs more testing

pkg/admission/validator/shoot.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Validate validates the given shoot objects.
+func (s *shoot) Validate(_ context.Context, newObj, oldObj client.Object) error {
+	shoot, ok := newObj.(*core.Shoot)
+	if !ok {
+		return fmt.Errorf("wrong object type %T", newObj)
+	}
+
+	if shoot.Spec.Provider.ControlPlaneConfig == nil {
+		return nil
```
breuerfelix: Why are we returning nil here? Shouldn't we error? The ske-api creates a controlplaneconfig at creation so it should never be nil and therefore error.

pkg/admission/validator/shoot.go
breuerfelix: maybe we should also create these skeletons for the controlplaneconfig aswell? Even though they might be empty, it indicates that we intentionally left them empty and it is not a bug, that we do not validate the update of the controlplaneconfig.

pkg/admission/validator/shoot.go
```diff
@@ -0,0 +1,68 @@
...
+			return fmt.Errorf("wrong object type %T for old object", oldObj)
+		}
+		oldInfraConfig, err := helper.InfrastructureConfigFromRawExtension(oldShoot.Spec.Provider.InfrastructureConfig)
+		if err != nil {
+			return err
+		}
+		allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfigUpdate(oldInfraConfig, infraConfig, field.NewPath("spec").Child("providerConfig").Child("infrastructureConfig"))...)
+	}
+
+	return allErrs.ToAggregate()
```
breuerfelix: as discussed, i would like to log the errors in the first version and always return nil.
After this changed is rolled out to prod, and we cannot see violations in loki in for existing shoots, we could actually make this validation failing

pkg/apis/stackit/validation/infrastructure.go
```diff
@@ -43,23 +34,38 @@ func ValidateInfrastructureConfig(infra *stackitv1alpha1.InfrastructureConfig, n
...
 
+	// validate that InfrastructureConfig.networks.id is a valid uuid
 	if infra.Networks.ID != nil {
 		if _, err := uuid.Parse(*infra.Networks.ID); err != nil {
-			allErrs = append(allErrs, field.Invalid(networksPath.Child("id"), infra.Networks.ID, "if network ID is provided it must be a valid OpenStack UUID"))
+			allErrs = append(allErrs, field.Invalid(networksPath.Child("id"), infra.Networks.ID, "if network ID is provided it must be a valid STACKIT Network ID"))
 		}
 	}
 
+	// eather InfrastructureConfig.networks.id or InfrastructureConfig.networks.worker(s) has to be set
```
breuerfelix:
```suggestion
	// either InfrastructureConfig.networks.id or InfrastructureConfig.networks.worker(s) has to be set
```

pkg/admission/validator/shoot.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	infraConfig, err := helper.InfrastructureConfigFromRawExtension(shoot.Spec.Provider.InfrastructureConfig)
+	if err != nil {
+		return err
+	}
+
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfig(cpConfig, shoot.Spec.Kubernetes.Version, field.NewPath("spec").Child("providerConfig"))...)
```
breuerfelix:
```suggestion
	allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfig(cpConfig, shoot.Spec.Kubernetes.Version, field.NewPath("spec").Child("provider").Child("controlPlaneConfig"))...)
```

pkg/admission/validator/shoot.go
```diff
@@ -0,0 +1,68 @@
...
+
+	allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfig(infraConfig, shoot.Spec.Networking.Nodes, field.NewPath("spec").Child("providerConfig").Child("infrastructureConfig"))...)
```
breuerfelix:
```suggestion
	allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfig(infraConfig, shoot.Spec.Networking.Nodes, field.NewPath("spec").Child("provider").Child("infrastructureConfig"))...)
```

pkg/admission/validator/shoot.go
breuerfelix:
```suggestion
		allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfigUpdate(oldInfraConfig, infraConfig, field.NewPath("spec").Child("provider").Child("infrastructureConfig"))...)
```

pkg/admission/validator/shoot.go
breuerfelix: `spec.Networking` can be nil and therefore `spec.Networking.Nodes` could panic

pkg/admission/validator/shoot.go
nschad: ~~Isn't it providerConfig? @breuerfelix I think you are mistaken.~~

Edit: NVM he is right

pkg/admission/validator/shoot.go
dergeberl: All fields are optional as some are defaulted. And others (not using ske-api) dont have to specify it in the beginning

breuerfelix: Then i would only skip the controleplaneconfig update if this field is nil since now you would also skip the infrastructure validation aswell, if the controlplaneconfig is nil

breuerfelix:
Just for the record: all current PROD shoots are not violating the current validation.

pkg/admission/validator/shoot.go
dergeberl: Got it, thanks!

pkg/admission/validator/shoot.go
dergeberl: Good catch, fixed and also added a test for it.

pkg/admission/validator/shoot.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			return err
+		}
+
+		allErrs = append(allErrs, stackitvalidation.ValidateInfrastructureConfigUpdate(oldInfraConfig, infraConfig, field.NewPath("spec").Child("provider").Child("infrastructureConfig"))...)
+		allErrs = append(allErrs, stackitvalidation.ValidateControlPlaneConfigUpdate(oldCpConfig, cpConfig, field.NewPath("spec").Child("provider").Child("controlPlaneConfig"))...)
+	}
+
+	// only log errors for now to be able to check the logs after a while before actually fail validation.
+	logger.Error(allErrs.ToAggregate(), "Shoot would have failed validation", "name", shoot.Name, "namespace", shoot.Namespace)
```
breuerfelix: this would log everytime, even for valid shoots.

rather use if allErrs.ToAggregate() != nil -> log

breuerfelix (approved):

breuerfelix:

dergeberl:
/unhold
was tested and we only do logging for now

PR #210, opened 2026-07-15
Add stricter validation for CSI compatibility mode

dergeberl:

/kind enhancement
/cc @ftl 

This PR adds more validation for `storage.CSI.CompatibilityMode`, so it only can be set to non default if CSI stackit is used.

Additionally this PR adds more tests for the rest of the validations.


breuerfelix (approved):

dergeberl:

PR #242, opened 2026-08-07
admission: remove PTP webhook for flatcar; remove `MutateDisableNTP` feature gate

nschad:

/hold
/kind cleanup

This webhook disabled ntpd for flatcar versions over >= 4230.2.1. The oldest flatcar version support we support is now 4230.2.3, therefore it is already enabled for all. This PR removes the webhook entirely as it's no longer needed.

Fixes #

**Breaking changes**:


pkg/admission/mutator/shoot.go
```diff
@@ -10,45 +9,33 @@ import (
...
 func NewShootMutator(mgr manager.Manager) extensionswebhook.Mutator {
-	logger.Info("MutateDisableNTP", "enabled", feature.Gate.Enabled(feature.MutateDisableNTP))
-	logger.Info("FlatcarImageVersion", "version", FlatcarImageVersion)
-
 	return &shoot{
 		decoder: serializer.NewCodecFactory(mgr.GetScheme(), serializer.EnableStrict).UniversalDecoder(),
 	}
 }
 
-func (s *shoot) Mutate(ctx context.Context, newObj, oldObj client.Object) error {
```
dergeberl, line 51: This Mutate does now basically nothing. I think it is not worth keeping it. Or do you have a use-case in mind that will come soon?

nschad: no you are right, we can drop it.

PR #250, opened 2026-08-11
Introduce new field `DNSServers` into `InfrastructureStatus`

nschad:

/kind enhancement

The `Subnet` field from the `InfrastructureStatus` will be removed, since the concept of a `Subnet` doesn't exist anymore in the new IaaS API. Additionally I also marked the **completely unused** fields of `ShareNetwork` as deprecated.

This new PR adds a new field called `DNSServer` of type `[]string` so we can save the `DNSServer` to the InfrastructureStatus, which is needed in the SKE internally developed cluster-error controller.

Fixes #

**Breaking changes**:

### DIFF
--- a/pkg/apis/stackit/v1alpha1/types_infrastructure.go
+++ b/pkg/apis/stackit/v1alpha1/types_infrastructure.go
@@ -43,6 +43,8 @@ type Networks struct {
 	SubnetID *string `json:"subnetId,omitempty"`
 	// ShareNetwork holds information about the share network (used for shared file systems like NFS)
 	// +optional
+	//
+	// Deprecated: Unused and will be removed later on
 	ShareNetwork *ShareNetwork `json:"shareNetwork,omitempty"`
 	// DNSServers overrides the default dns configuration from cloud profile
 	// +optional
@@ -56,6 +58,8 @@ type Router struct {
 }
 
 // ShareNetwork holds information about the share network (used for shared file systems like NFS)
+//
+// Deprecated: Unused and will be removed later on
 type ShareNetwork struct {
 	// Enabled is the switch to enable the creation of a share network
 	Enabled bool `json:"enabled"`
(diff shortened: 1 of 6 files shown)


pkg/apis/stackit/v1alpha1/types_infrastructure.go
```diff
@@ -90,10 +94,17 @@ type NetworkStatus struct {
 	FloatingPool FloatingPoolStatus `json:"floatingPool"`
 	// Router contains information about the Router and related resources.
 	Router RouterStatus `json:"router"`
+	// DNSServer contains the networks configured dnsServers
+	// +optional
+	DNSServers *[]string `json:"dnsServers,omitempty"`
 	// Subnets is a list of subnets that have been created.
+	//
+	// Deprecated: Will be removed once fully migrated to IaaS API
```
dergeberl:
```suggestion
	// Deprecated: OpenStack-only; not used for STACKIT.
```

pkg/controller/infrastructure/stackit/infraflow/context.go
```diff
@@ -180,7 +180,13 @@ func (fctx *FlowContext) computeInfrastructureStatus() *stackitv1alpha1.Infrastr
 
 	status.Node.KeyName = ptr.Deref(fctx.state.Get(NameKeyPair), "")
 
+	if fctx.dnsNameservers != nil {
+		status.Networks.DNSServers = fctx.dnsNameservers
+	}
+
+	// TODO: Remove once migrated fully to IaaS API
 	if v := fctx.state.Get(IdentifierSubnet); v != nil {
+		//nolint:staticcheck // SA1019: Remove once cluster-controller does not use that field anymore
```
dergeberl: This is also used by the openstack MCM so this can not be removed until we have migrated everything. This can not be removed after the DNS checks in cluster-controller are moved

nschad: Well it does say Remove one migrated fully to IaaS API, this includes removal of OpenStack.

nschad: I've tried clarifying it. Is it fine now?

PR #251, opened 2026-08-12
Prepare workload identity federation for integration tests

MichaelEischer:

/kind enhancement

Prepare integration tests to allow injecting the service account credentials for the project creation using workload identity federation. As the actual test should use the service account created by the test wrapper, overwrite the environment [variables checked by the STACKIT SDK](https://github.com/stackitcloud/stackit-sdk-go/blob/401fa8f1bd2f202b6668eb0c99f17bd9bcd770de/core/clients/workload_identity_flow.go#L165) such that the workload identity is no longer detected.

Fixes #

**Breaking changes**:

### DIFF
--- a/test/project-wrapper/main.go
+++ b/test/project-wrapper/main.go
@@ -11,6 +11,8 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"slices"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/project-wrapper/main.go
```diff
@@ -91,7 +93,13 @@ func run() error {
...
 	// is limited in our CI tooling. Therefore, we can ignore it.
 	cmd := exec.CommandContext(ctx, os.Args[1], os.Args[2:]...) // #nosec G204 G702
-	cmd.Env = append(os.Environ(),
+	cmd.Env = append(
+		// remove `STACKIT_SERVICE_ACCOUNT_EMAIL` to prevent using workflow identity path
+		slices.DeleteFunc(os.Environ(), func(env string) bool {
+			return strings.HasPrefix(env, "STACKIT_SERVICE_ACCOUNT_EMAIL=")
+		}),
+		// also shadow the federated token file
+		"STACKIT_FEDERATED_TOKEN_FILE=/dev/null",
```
timebertt, line 102: I would expect that this will fail here: https://github.com/stackitcloud/stackit-sdk-go/blob/401fa8f1bd2f202b6668eb0c99f17bd9bcd770de/core/oidcadapters/filesystem.go#L21-L23
But we can try

MichaelEischer: Yes, this will lead to an error in https://github.com/stackitcloud/stackit-sdk-go/blob/401fa8f1bd2f202b6668eb0c99f17bd9bcd770de/core/clients/workload_identity_flow.go#L173 . And that error will eventually trigger falling back to key auth at https://github.com/stackitcloud/stackit-sdk-go/blob/c08635aecd68fb9e075830f39e245402092a9683/core/auth/auth.go#L97

timebertt: Understood, thanks :) 
/lgtm

breuerfelix (approved):

PR #252, opened 2026-08-12
chore: Increase --machine-creation-timout to 45m

viragvoros:
/kind bug

To reduce load on openstack we need to allow machines to stay in creation for a longer time. Increasing the `--machine-creation-timout` to 45m results in a higher chance of getting the creation through.


nschad:
This is really not a bug!!
/approve

PR #272, opened 2026-08-26
migrate mcm-provider from openstack to stackit

aniruddha2000:

/kind enhancement

This PR replaces mcm provider from openstack to stackit. Replaces openstack provider ID with stackit provider ID and add `stackit.cloud/migrated-machine: true` annotation to machines and puts annotation `stackit.cloud/machine-controller-manager-migrated: true` to workers once the migration is done.

Fixes #

**Breaking changes**:

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -77,6 +77,7 @@ func (w *workerDelegate) DeployMachineClasses(ctx context.Context) error {
 	}
 
 	if feature.MigrateStackitMachineControllerManager(w.cluster) && w.worker.Annotations[workerMigratedAnnotation] != "true" {
+		fmt.Println("Migrating Stackit Machine Controller Manager to Gardener Machine Controller Manager...")
 		err = w.migrateMachines(ctx)
 		if err != nil {
 			return err
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -74,7 +74,7 @@ func (w *workerDelegate) DeployMachineClasses(ctx context.Context) error {
 		return err
 	}
 
-	if feature.MigrateStackitMachineControllerManager(w.cluster) {
+	if feature.MigrateStackitMachineControllerManager(w.cluster) && w.worker.Annotations[workerMigratedAnnotation] != "true" {
 		err = w.migrateMachines(ctx)
 		if err != nil {
 			return err
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -76,8 +76,9 @@ func (w *workerDelegate) DeployMachineClasses(ctx context.Context) error {
 		return err
 	}
 
+	fmt.Println("Deploying OpenStack machine classes...")
 	if feature.MigrateStackitMachineControllerManager(w.cluster) && w.worker.Annotations[workerMigratedAnnotation] != "true" {
-		fmt.Println("Migrating Stackit Machine Controller Manager to Gardener Machine Controller Manager...")
+		fmt.Println("Migrating Stackit Machine Controller Manager to Gardener Machine Controller Manager...") // TODO: remove later
 		err = w.migrateMachines(ctx)
 		if err != nil {
 			return err
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -516,14 +516,20 @@ func (w *workerDelegate) markWorkerAsMigrated(ctx context.Context) error {
 
 func serverIDFromProviderID(providerID string) (string, error) {
 	patterns := []*regexp.Regexp{
-		regexp.MustCompile(`^openstack:///[^/]+/([^/]+)$`),
-		regexp.MustCompile(`^stackit://[^/]+/([^/]+)$`),
+		regexp.MustCompile(`^openstack:///[^/]+/(?P<serverID>[^/]+)$`),
+		regexp.MustCompile(`^stackit://[^/]+/(?P<serverID>[^/]+)$`),
 	}
 
 	for _, pattern := range patterns {
-		matches := pattern.FindStringSubmatch(providerID)
-		if len(matches) == 2 {
-			return matches[1], nil
+		match := pattern.FindStringSubmatch(providerID)
+		if len(match) == 0 {
+			continue
+		}
+
+		for i, name := range pattern.SubexpNames() {
+			if name == "serverID" {
+				return match[i], nil
+			}
 		}
 	}
 
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -40,6 +40,7 @@ import (
 
 const shouldMigrateMachineAnnotation = "stackit.cloud/machine-should-be-migrated"
 const migratedMachineAnnotation = "stackit.cloud/migrated-machine"
+const workerMigratedAnnotation = "stackit.cloud/machine-controller-manager-migrated"
 
 // MachineClassKind yields the name of the machine class kind used by OpenStack provider.
 func (w *workerDelegate) MachineClassKind() string {
@@ -495,5 +496,17 @@ func (w *workerDelegate) migrateMachines(ctx context.Context) error {
 		}
 	}
 
-	return nil
+	return w.markWorkerAsMigrated(ctx)
+}
+
+func (w *workerDelegate) markWorkerAsMigrated(ctx context.Context) error {
+	patchWorker := client.MergeFrom(w.worker.DeepCopy())
+
+	if w.worker.Annotations == nil {
+		w.worker.Annotations = make(map[string]string)
+	}
+
+	w.worker.Annotations[workerMigratedAnnotation] = "true"
+
+	return w.seedClient.Patch(ctx, w.worker, patchWorker)
 }
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
@@ -34,7 +35,11 @@ import (
 	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/openstack"
 	"github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/stackit"
 	stackitutils "github.com/stackitcloud/gardener-extension-provider-stackit/v2/pkg/utils"
+	iaas2 "github.com/stackitcloud/stackit-sdk-go/services/iaas/v2api"
 )
+
+const shouldMigrateMachineAnnotation = "stackit.cloud/machine-should-be-migrated"
+const migratedMachineAnnotation = "stackit.cloud/migrated-machine"
 
 // MachineClassKind yields the name of the machine class kind used by OpenStack provider.
 func (w *workerDelegate) MachineClassKind() string {
@@ -63,7 +68,19 @@ func (w *workerDelegate) DeployMachineClasses(ctx context.Context) error {
 	if feature.UseStackitMachineControllerManager(w.cluster) {
 		chartPath = "machineclass-stackit"
 	}
-	return w.seedChartApplier.ApplyFromEmbeddedFS(ctx, charts.InternalChart, filepath.Join(charts.InternalChartsPath, chartPath), w.worker.Namespace, "machineclass", kubernetes.Values(map[string]any{"machineClasses": w.machineClasses}))
+	err := w.seedChartApplier.ApplyFromEmbeddedFS(ctx, charts.InternalChart, filepath.Join(charts.InternalChartsPath, chartPath), w.worker.Namespace, "machineclass", kubernetes.Values(map[string]any{"machineClasses": w.machineClasses}))
+	if err != nil {
+		return err
+	}
+
+	if feature.MigrateStackitMachineControllerManager(w.cluster) {
+		err = w.migrateMachines(ctx)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // GenerateMachineDeployments generates the configuration for the desired machine deployments.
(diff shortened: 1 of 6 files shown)


pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,92 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+		m.Annotations[shouldMigrateMachineAnnotation] = "true"
+		m.Annotations[migratedMachineAnnotation] = "true"
+		err = w.seedClient.Patch(ctx, &m, patchAnnotations)
+		if err != nil {
+			return err
+		}
+
+		if m.Spec.ProviderID != "" {
+			providerIDParts := strings.Split(m.Spec.ProviderID, "/")
+			if len(providerIDParts) == 0 {
```
nschad: unreachable

https://pkg.go.dev/strings#Split

Len is always atleast "1"

nschad: Also why do you split on just "/", shouldn't this be "://"?

pkg/controller/worker/machines.go
```diff
@@ -24,6 +25,7 @@ import (
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	gardenutils "github.com/gardener/gardener/pkg/utils"
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+	iaas2 "github.com/stackitcloud/stackit-sdk-go/services/iaas/v2api"
```
nschad:
```suggestion
	iaas "github.com/stackitcloud/stackit-sdk-go/services/iaas/v2api"
```

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,92 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+			patch := client.MergeFrom(m.DeepCopy())
+			m.Spec.ProviderID = fmt.Sprintf("stackit://%s/%s", iaas.ProjectID(), serverID)
+			err = w.seedClient.Patch(ctx, &m, patch)
+			if err != nil {
+				return err
+			}
+
+			_, err = iaas.UpdateServer(ctx, serverID, iaas2.UpdateServerPayload{
+				Labels: map[string]any{
+					//	// TODO refine labels
```
nschad: What about the TODO?

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,92 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+		if err != nil {
+			return err
+		}
+
+		if m.Spec.ProviderID != "" {
+			providerIDParts := strings.Split(m.Spec.ProviderID, "/")
+			if len(providerIDParts) == 0 {
+				return fmt.Errorf("migrateMachines: malformed machine provider ID: %s", m.Spec.ProviderID)
+			}
+			serverID := providerIDParts[len(providerIDParts)-1]
```
nschad: fine but hard to read, either add a comment or improve it

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,92 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+
+			patch := client.MergeFrom(m.DeepCopy())
+			m.Spec.ProviderID = fmt.Sprintf("stackit://%s/%s", iaas.ProjectID(), serverID)
```
nschad: do we not have a constant for `stackit://` ?

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,92 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+
+	if len(migrateMachines) == 0 {
+		// no old openstack machine
+		return w.markWorkerAsMigrated(ctx)
+	}
+
+	iaas, err := w.stackitClient.IaaS(ctx, w.seedClient, w.worker.Spec.SecretRef)
+	if err != nil {
+		return err
+	}
```
nschad: does this authenticate the client every time the function is called? Could we do this somewhere else?

pkg/controller/worker/machines.go
nschad: suggestion check with regex pattern for correct format.

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,104 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
 }
+
+func (w *workerDelegate) migrateMachines(ctx context.Context) error {
+	var (
+		allMachines     machinev1alpha1.MachineList
+		migrateMachines []machinev1alpha1.Machine
+	)
+	var (
+		openStackProviderIDPattern = regexp.MustCompile(`^openstack:///[^/]+/([^/]+)$`)
+		stackitProviderIDPattern   = regexp.MustCompile(`^stackit://[^/]+/([^/]+)$`)
```
nschad: no this should be within func init() so it only runs once

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,104 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+	)
+
+	const stackitProviderID = "stackit://"
```
nschad:
```suggestion
```

Also misplaced, either find an existing const file or top of the file.

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,104 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+				// A retry can resume, after the provider ID was already converted but
+				// before updating the server labels completed.
+				matches = stackitProviderIDPattern.FindStringSubmatch(m.Spec.ProviderID)
+			}
+			if len(matches) != 2 {
+				return fmt.Errorf(
+					"migrateMachines: malformed machine provider ID: %s",
+					m.Spec.ProviderID,
+				)
+			}
```
nschad: refactor into helper function

pkg/controller/worker/machines.go
```diff
@@ -395,3 +415,104 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+			// capture server ID from provider ID
+			serverID := matches[1]
```
nschad: somehow this is even more confusing. Can we do named capture groups in the regex?

pkg/controller/worker/machines.go
nschad: Fix the comment atleast @aniruddha2000

nschad (changes requested):

pkg/controller/worker/machines.go
aniruddha2000: I dont get it, do you mean something like regexp.MustCompile(`^openstack:///[^/]+/(?P<serverID>[^/]+)$`) ?

pkg/controller/worker/machines.go
aniruddha2000: I dont get the init() func, I have removed this vars and put them in a helper func all together.

pkg/controller/worker/machines.go
nschad: Yes

pkg/controller/worker/machines.go
```diff
@@ -516,14 +516,20 @@ func (w *workerDelegate) markWorkerAsMigrated(ctx context.Context) error {
 
 func serverIDFromProviderID(providerID string) (string, error) {
 	patterns := []*regexp.Regexp{
-		regexp.MustCompile(`^openstack:///[^/]+/([^/]+)$`),
-		regexp.MustCompile(`^stackit://[^/]+/([^/]+)$`),
+		regexp.MustCompile(`^openstack:///[^/]+/(?P<serverID>[^/]+)$`),
+		regexp.MustCompile(`^stackit://[^/]+/(?P<serverID>[^/]+)$`),
```
nschad, line 524: move `regexp.MustCompile` into `func init()`

pkg/controller/worker/machines.go
```diff
@@ -406,3 +428,116 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+		err = w.seedClient.Patch(ctx, &m, patchAnnotations)
+		if err != nil {
+			return err
+		}
+
+		if m.Spec.ProviderID != "" {
+			serverID, err := serverIDFromProviderID(m.Spec.ProviderID)
+			if err != nil {
+				return fmt.Errorf("migrateMachines: %w", err)
+			}
```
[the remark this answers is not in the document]
aniruddha2000: skipping this, not sure if this will affect the migration somehow!

pkg/controller/worker/machines.go
```diff
@@ -406,3 +428,116 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+				return err
+			}
+
+			_, err = w.iaaSClient.UpdateServer(ctx, serverID, iaas.UpdateServerPayload{
+				Labels: map[string]any{
+					// TODO refine labels
+					"mcm.gardener.cloud/machine":      m.Name,
+					"mcm.gardener.cloud/machineclass": m.Spec.Class.Name,
+					"mcm.gardener.cloud/role":         "node",
+				},
```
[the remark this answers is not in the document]
aniruddha2000: Refine labels for later purposes

pkg/controller/worker/machines.go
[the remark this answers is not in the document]
breuerfelix: This should be checked, this is a valid comment. We should handle an empty provider ID, since machines could be in state creating while doing the migration.

pkg/controller/worker/machines.go
```diff
@@ -537,6 +537,10 @@ func serverIDFromProviderID(providerID string) (string, error) {
 				return match[i], nil
 			}
 		}
+
+		if idx := pattern.SubexpIndex("serverID"); idx >= 0 && idx < len(match) {
```
breuerfelix, line 542: The AI meant to use this instead of the lines above, not additionally.

aniruddha2000: ahh i have this exactly in the local, forgot to push

pkg/controller/worker/machines.go
```diff
@@ -465,30 +465,32 @@ func (w *workerDelegate) migrateMachines(ctx context.Context) error {
 			return err
 		}
 
-		if m.Spec.ProviderID != "" {
-			serverID, err := serverIDFromProviderID(m.Spec.ProviderID)
-			if err != nil {
-				return fmt.Errorf("migrateMachines: %w", err)
-			}
+		if m.Spec.ProviderID == "" {
```
breuerfelix: we cannot just error out here.
we have to answer the following questions:
can the provider id even be empty? if yes, in which cases? what is the state of the server if the id is empty?
what happens when we return an error here if the id is empty, this will result in a "failed migration" right? is this the expected behaviour?

pkg/controller/worker/machines.go
```diff
@@ -63,7 +73,19 @@ func (w *workerDelegate) DeployMachineClasses(ctx context.Context) error {
...
+	err := w.seedChartApplier.ApplyFromEmbeddedFS(ctx, charts.InternalChart, filepath.Join(charts.InternalChartsPath, chartPath), w.worker.Namespace, "machineclass", kubernetes.Values(map[string]any{"machineClasses": w.machineClasses}))
+	if err != nil {
+		return err
+	}
+
+	if feature.MigrateStackitMachineControllerManager(w.cluster) && w.worker.Annotations[workerMigratedAnnotation] != "true" {
+		err = w.migrateMachines(ctx)
+		if err != nil {
+			return err
+		}
```
breuerfelix:
```suggestion
		if err := w.migrateMachines(ctx); err != nil {
			return err
		}
```

pkg/controller/worker/machines.go
breuerfelix: @aniruddha2000 why have you resolved this comment?
as discussed, we should no error out here, just remove these 3 lines. 
if the providerID is empty, the machine will get cleaned up by the mcm anyways after some time

pkg/controller/worker/machines.go
```diff
@@ -406,3 +428,118 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+		err = w.seedClient.Patch(ctx, &m, patchAnnotations)
+		if err != nil {
+			return err
+		}
+
+		if m.Spec.ProviderID == "" {
+			return fmt.Errorf("cannot migrate machine %s: providerID is empty (provisioning in progress)", m.Name)
+		}
+
+		if m.Spec.ProviderID != "" {
```
breuerfelix, line 475: add a comment above on why empty providerID is also fine (this is stated in the ADR), so the next person who reads that code knows that we intentionally skip them

pkg/controller/worker/machines.go
```diff
@@ -406,3 +428,118 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+		return w.markWorkerAsMigrated(ctx)
+	}
+
+	for _, m := range migrateMachines {
+		patchAnnotations := client.MergeFrom(m.DeepCopy())
+		if m.Annotations == nil {
+			m.Annotations = make(map[string]string)
+		}
+		m.Annotations[shouldMigrateMachineAnnotation] = "true"
+		m.Annotations[migratedMachineAnnotation] = "true"
```
breuerfelix, line 463: add a comment what this annotation does, and that we need it for cleaning up the nics (stated in the ADR)

pkg/controller/worker/machines.go
```diff
@@ -406,3 +428,118 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+			if err != nil {
+				return err
+			}
+
+			_, err = w.iaaSClient.UpdateServer(ctx, serverID, iaas.UpdateServerPayload{
+				Labels: map[string]any{
+					// TODO refine labels
+					"mcm.gardener.cloud/machine":      m.Name,
+					"mcm.gardener.cloud/machineclass": m.Spec.Class.Name,
+					"mcm.gardener.cloud/role":         "node",
```
breuerfelix:
```suggestion
					"kubernetes.io/machine":      m.Name,
					"kubernetes.io/machineclass": m.Spec.Class.Name,
					"kubernetes.io/role":         "node",
```

breuerfelix: AFAIK these are the labels that the stackit-mcm are using.
double check that, when using the stackit mcm, creating a new machine, what labels are present on this machine?

pkg/controller/worker/machines.go
```diff
@@ -406,3 +428,118 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+)
+
+func init() {
+	providerIDPatterns = []*regexp.Regexp{
+		regexp.MustCompile(`^openstack:///[^/]+/(?P<serverID>[^/]+)$`),
+		regexp.MustCompile(`^stackit://[^/]+/(?P<serverID>[^/]+)$`),
+	}
+}
+
+func serverIDFromProviderID(providerID string) (string, error) {
```
breuerfelix, line 535: add a unit test for this function

pkg/controller/worker/machines.go
aniruddha2000: :) whats wrong with my git, i deleted it and pushed a commit bellow this comment, but somehow it did not! sorrry! :( 

commit also says "revert provider id is empty review"

pkg/controller/worker/machines.go
aniruddha2000: skipping this as its private func

pkg/controller/worker/machines.go
```diff
@@ -406,3 +427,122 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+		m.Annotations[migratedMachineAnnotation] = "true"
+		err = w.seedClient.Patch(ctx, &m, patchAnnotations)
+		if err != nil {
+			return err
+		}
+
+		// It is okay to skip machine without a provider ID, as there is a fallback to get
+		// the server by name in case there is no providerID during deletion by the MCM.
+		// Normally this is done with a label containing the machine name and a label selector.
+		// In case of a migrated machine with the stackit.cloud/migrated-machine annotation the deletion needs
```
breuerfelix: this is wrong, since we have no provider id, we cannot update the labels.
in case the provider id is missing, we will fetch all servers and filter by name

pkg/controller/worker/machines.go
```diff
@@ -406,3 +427,122 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+	}
+
+	for _, m := range migrateMachines {
+		patchAnnotations := client.MergeFrom(m.DeepCopy())
+		if m.Annotations == nil {
+			m.Annotations = make(map[string]string)
+		}
+		// This annotation is deleted when the server is updated, otherwise its incomplete migrated machine.
+		m.Annotations[shouldMigrateMachineAnnotation] = "true"
+		// The MCM needs to get and delete the NICs of the machine, as they were created separately which needs dedicated deletion
```
breuerfelix:
```suggestion
		// The MCM needs to get and delete the NICs of the machine, as they were created separately which needs dedicated deletion. The MCM used this annotation as a marker for the said NIC cleanup.
```

pkg/controller/worker/machines.go
```diff
@@ -406,3 +427,122 @@ func EnsureUniformMachineImages(images []stackitv1alpha1.MachineImage, definitio
...
+			patch := client.MergeFrom(m.DeepCopy())
+			m.Spec.ProviderID = fmt.Sprintf("%s%s/%s", stackitProviderID, w.iaaSClient.ProjectID(), serverID)
+			err = w.seedClient.Patch(ctx, &m, patch)
+			if err != nil {
+				return err
+			}
+
+			_, err = w.iaaSClient.UpdateServer(ctx, serverID, iaas.UpdateServerPayload{
+				Labels: map[string]any{
+					// TODO refine labels
```
breuerfelix, line 487: you can remove this TODO now

aniruddha2000: skiping this for now.

PR #212, opened 2026-07-15
Enable shoot validation in addmission webhook

dergeberl:
**How to categorize this PR?**

/kind enhancement

**What this PR does / why we need it**:

With PR #203 we added the shoot object to the validation webhook in logging mode but did not fail the validation. This PR enables the actual validation.

PR #281, opened 2026-09-03
Deprecate OpenStack only fields in InfrastructureConfig

dergeberl:

/kind enhancement

This is a followup from #280.

The `floatingPoolName` was required but never used in a STACKIT only Cluster (since the validation is in place #212). It is only used in the STACKIT Infra-Controller in case there are OpenStack Credentials to pull the IDs of it in the Status.

The #280 PR removed the validation. This PR removes that the FloatingNetID is get via OpenStack in the STACKIT Infra-Controller (even if there are OpenStack Credenitals), as there is no need for it. This PR sets all only OpenStack fields in the Infrastructure to Depreciated. In OpenStack we save the EgressIPs in the provider status of the Infrastructure object to put it later in the Infrastructure Object. This is now done directly via context for STACKIT to not use the unneeded deprecated field in STACKIT Infra Controller.

Fixes #

**Breaking changes**:

### DIFF
--- a/pkg/apis/stackit/v1alpha1/types_infrastructure.go
+++ b/pkg/apis/stackit/v1alpha1/types_infrastructure.go
@@ -15,9 +15,14 @@ import (
 type InfrastructureConfig struct {
 	metav1.TypeMeta `json:",inline"`
 	// FloatingPoolName contains the FloatingPoolName name in which LoadBalancer FIPs should be created.
+	//
+	// Deprecated: OpenStack-only; not used for STACKIT.
+	// +optional
 	FloatingPoolName string `json:"floatingPoolName"`
 	// FloatingPoolSubnetName contains the fixed name of subnet or matching name pattern for subnet
 	// in the Floating IP Pool where the router should be attached to.
+	//
+	// Deprecated: OpenStack-only; not used for STACKIT.
 	// +optional
 	FloatingPoolSubnetName *string `json:"floatingPoolSubnetName,omitempty"`
 	// Networks is the OpenStack specific network configuration
@@ -27,6 +32,8 @@ type InfrastructureConfig struct {
 // Networks holds information about the Kubernetes and infrastructure networks.
 type Networks struct {
 	// Router indicates whether to use an existing router or create a new one.
+	//
+	// Deprecated: OpenStack-only; not used for STACKIT.
 	// +optional
 	Router *Router `json:"router,omitempty"`
 	// Worker is a CIDRs of a worker subnet (private) to create (used for the VMs).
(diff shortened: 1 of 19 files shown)


dergeberl:
/hold
needs testing in for different scenarios:
- Openstack Infra-controller
- STACKIT Infra-Controller
- STACKIT Infra-Controller (without OpenStack credentials)

pkg/apis/stackit/v1alpha1/types_infrastructure.go
```diff
@@ -91,8 +100,12 @@ type NetworkStatus struct {
 	// Name is the Network name.
 	Name string `json:"name"`
 	// FloatingPool contains information about the floating pool.
+	//
+	// Deprecated: OpenStack-only; not used for STACKIT.
 	FloatingPool FloatingPoolStatus `json:"floatingPool"`
 	// Router contains information about the Router and related resources.
+	//
+	// Deprecated: OpenStack-only; not used for STACKIT.
```
nschad: This is used, also for STACKIT. Especially the `ExternalFixedIPs` or so is used in other extensions

pkg/apis/stackit/v1alpha1/types_infrastructure.go
```diff
@@ -109,30 +122,48 @@ type NetworkStatus struct {
...
 	ID string `json:"id"`
 	// IP is the router ip.
 	//
-	// Deprecated: use ExternalFixedIPs instead.
+	// Deprecated: use ExternalFixedIPs instead. OpenStack-only; not used for STACKIT.
 	IP string `json:"ip"`
 	// ExternalFixedIPs is the list of the router's assigned external fixed IPs.
+	//
+	// Deprecated: OpenStack-only; not used for STACKIT.
```
nschad, line 138: Isn't this used in other extensions?

nschad (changes requested):

pkg/apis/stackit/v1alpha1/types_infrastructure.go
dergeberl: Other extensions should depend on the `infra.Status.EgressCIDRs` (They should not depend on the the provider part). Do you have anything special in mind?

Example for acl: 
https://github.com/stackitcloud/gardener-extension-acl/blob/33491918ea575b3241d4c594d80ea517d9d80670/pkg/controller/actuator.go#L191

nschad:
> /hold needs testing in for different scenarios:
> 
> * Openstack Infra-controller
> * STACKIT Infra-Controller
> * STACKIT Infra-Controller (without OpenStack credentials)

I've tested all cases, lgtm

For the letter i created a extra `Cloudprofile` + new secret with STACKIT only and dropped floatingPool fields

/lgtm

nschad (approved):

PR #300, opened 2026-09-17
Implement Mutating Webhook to Block NTP in Ubuntu 26.04

viragvoros:
/kind enhancement

Ubuntu 26.04 does not have NTP. As `gardener-extension-os-ubuntu` [does not allow to turn off NTP](https://github.com/gardener/gardener-extension-os-ubuntu/blob/0115f0ecf675aad73346f88e51b83ba655f532be/pkg/controller/config/v1alpha1/types.go#L115), we need a temporary fix in `gardener-extension-provider-stackit`. Long term, NTPSEC will be introduced upstream.

This is a temporary fix, after upstream implementation, it can be reverted.


nschad:
Won't be implemented here.

